PR #9, opened 2020-01-15
Recipe for a conda package

raimis:
- [x] Recipe for a conda package
- [ ] Building documentation

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,0 +1,57 @@
+{% set name = "openmm-torch" %}
+{% set version = "0.0" %}
+{% set build = 0 %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  path: ..
+
+build:
+  number: {{ build }}
+  rpaths:
+    - lib/
+    # Note: $PY_VER isn't expanded here
+    - lib/python3.6/site-packages/torch/lib/ # [py==36]
+    - lib/python3.7/site-packages/torch/lib/ # [py==37]
+    - lib/python3.8/site-packages/torch/lib/ # [py==38]
+  script:
+    - cmake $SRC_DIR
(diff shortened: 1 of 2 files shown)


raimis:
There are two problems:
- OpenMM API has changed. The OpenMM 7.4.1 package from the `omnia` channel doesn't work jet, but `omnia-dev` has newer one.
- OpenMM packages (both from `omnia` and `omnia-dev`) are compiled with the old C++ ABI (https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html), but PyTorch packages (both from `default` and `conda-forge` channels) are compiled with the new C++ ABI.

raimis:
@peastman Is any reason why the old C++ ABI is used for the OpenMM packages?

peastman:
Conda packages for omnia projects are managed through the https://github.com/omnia-md/conda-recipes and https://github.com/omnia-md/conda-recipes-cuda repositories.  But we aren't yet at the point where we can build conda packages for this.  There are issues related to PyTorch that need to be resolved first, especially the fact that LibTorch isn't conda installable.  Once those issues are resolved we'll create conda packages using the omnia infrastructure.

raimis:
LibTorch is a PyTorch back-end and it comes with `pytorch` package (`conda install pytorch`).
Just its location i is non-standard (`$CONDA_PREFIX/lib/python3.7/site-packages/torch/`), but it contains headers, `libtorch.so`, and other depended libraries.

The only real problem, as mentioned above, is the C++ ABI incompatibility.

peastman:
> LibTorch is a PyTorch back-end and it comes with pytorch package

I wasn't aware of that.  Thanks!  I'll see if I can get the plugin to work with it.  If so, that will make things a lot easier.  That ABI incompatibility could be a problem though.

PR #15, opened 2020-07-31
Avoid race condition while copying forces on GPU

peastman:
Fixes #13.

PR #17, opened 2020-09-30
[BUGFIX] Synchronize CUDA context before force copying

yaoyic:
Pull request for fixing a bug due to CUDA synchronizing issues in the current code base.

### Bug Description
During our application of this plugin on a system with 350 particles for incorporating a neural network force (on CUDA platform), there is a strange behavior similar to the description of Issue #13, i.e., the average temperature is lower under the Langevin thermostat setting.

The output force from the plugin is sometimes correct, but most of the time very wrong (it has less than half of the expected magnitude) comparing to a direct evaluation of the neural network in PyTorch.

### Diagnosis

Pull #15 and #16 do not solve the problem on our side.

Line 85 of `platforms/cuda/src/CudaTorchKernels.cpp`: `energyTensor.backward()` *seems* async (at least for PyTorch version 1.5), although there's no proper description in the [official documents for C++ API](https://pytorch.org/cppdocs/api/classat_1_1_tensor.html?highlight=tensor#_CPPv4NK2at6Tensor8backwardERKN3c108optionalI6TensorEEN3c108optionalIbEEb).
Line 87 of `platforms/cuda/src/CudaTorchKernels.cpp`: `posTensor.grad()[.clone()]` returns (a reference of) the grad tensor without blocking until gradient calculation is actually finished.
Therefore, I suspect that the neural network operations is not properly finished.
Note that in the Python [documentation](https://pytorch.org/docs/stable/notes/cuda.html#asynchronous-execution), they do say that PyTorch operations on CUDA are **asynchronous**.

### Bug Fix
Synchronize with GPU before changing the CUDA context or executing the `addForceKernel`, so as to make sure that all operations on (Py)Torch side is complete.
However, somehow the C++ API itself doesn't expose an equivalent to [`torch.cuda.synchronize()`](https://pytorch.org/docs/stable/cuda.html#torch.cuda.synchronize) (at least according to their documentation).
I propose to call an equivalent function directly from the CUDA runtime APIs instead, which is used inside [some other code](https://github.com/pytorch/pytorch/blob/4aca63d38aaed749a72b31eab098f7305ea711f4/test/cpp/tensorexpr/test_cuda.cpp#L74) of PyTorch itself.
Current solution solves our problem.

### DIFF
--- a/platforms/cuda/src/CudaTorchKernels.cpp
+++ b/platforms/cuda/src/CudaTorchKernels.cpp
@@ -33,6 +33,7 @@
 #include "CudaTorchKernelSources.h"
 #include "openmm/internal/ContextImpl.h"
 #include <map>
+#include <cuda_runtime_api.h>
 
 using namespace TorchPlugin;
 using namespace OpenMM;
@@ -85,6 +86,8 @@ double CudaCalcTorchForceKernel::execute(ContextImpl& context, bool includeForce
         energyTensor.backward();
         // Note: "forceTensor" needs to be cloned due to a shared context (https://github.com/openmm/openmm-torch/issues/13)
         torch::Tensor forceTensor = posTensor.grad().clone();
+        // make sure that all calculations on PyTorch side is properly finished before changing CUDA context or starting the `addForcesKernel` of this plugin
+        cudaDeviceSynchronize();
         cu.setAsCurrent();
         void* data;
         if (cu.getUseDoublePrecision()) {
--- a/platforms/cuda/src/CudaTorchKernels.cpp
+++ b/platforms/cuda/src/CudaTorchKernels.cpp
@@ -90,6 +90,8 @@ double CudaCalcTorchForceKernel::execute(ContextImpl& context, bool includeForce
     vector<torch::jit::IValue> inputs = {posTensor};
     if (usePeriodic)
         inputs.push_back(boxTensor);
+    // synchronizing the current context before switching to PyTorch
+    CHECK_RESULT(cuCtxSynchronize(), "Error synchronizing CUDA context");
     torch::Tensor energyTensor = module.forward(inputs).toTensor();
     if (includeForces) {
         energyTensor.backward();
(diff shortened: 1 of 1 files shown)


raimis:
@yaoyic thanks for working on this.

OpenMM and PyTorch use the CUDA contexts not in the most compatible way (https://github.com/openmm/openmm-torch/issues/13#issuecomment-667999652, https://github.com/openmm/openmm-torch/issues/13#issuecomment-669289989). I was looking how to fix the context problem on the OpenMM side, but not yet solved that.

Normally OpenMM uses two CUDA streams (one for the reciprocal space PME, and the other for the rest). Calling `cudaDeviceSynchronize()` will synchronise both streams and impact overall performance. 

Preferably, you should use `cuCtxSynchronize`. @peastman tried that (https://github.com/openmm/openmm-torch/issues/13#issuecomment-667323570), but it was before the shared tensor problem (https://github.com/openmm/openmm-torch/pull/16) was fixed.

yaoyic:
@raimis thanks for your comments.

Yes, you are right! For complex cases (e.g., multiple simulations, other software, etc. running on the same GPU) `cuCtxSynchronize` is a more specific and also a correct solution. I tested to change `cudaDeviceSynchronize()` to `cuCtxSynchronize()` and the latter also solves the problem on my side.

I also included checking for the return value of synchronization, following an example in the [OpenMM code](https://github.com/openmm/openmm/blob/edbc8407b0361ce5152bad4bd91597080dee3b58/platforms/cuda/src/CudaParallelKernels.cpp#L107).

raimis:
Thanks @yaoyic! I'll try the fix with a few crashing system.

raimis:
@yaoyic I have run several simulations still there were crashes. After digging deeper, I found that a synchronisation is needed not only after the PyTorch execution, but also before. So, there must a synchronisation before each CUDA context switch.

raimis:
Could you add
```c++
CHECK_RESULT(cuCtxSynchronize(), "Error synchronizing CUDA context");
```
before
```c++
torch::Tensor energyTensor = module.forward(inputs).toTensor();
```
and verify that it works for you.

yaoyic:
@raimis I added the line of code you mentioned.

raimis:
@peastman I think, this is good to merge. Eventually, the context management has to be fixed to use the same context, but meanwhile this solves the problem.

peastman:
Thanks!

PR #23, opened 2021-01-19
Update README to be more comprehensible

jchodera:
@peastman: This PR updates the README to be more lucid in terms of how to install and use the `TorchForce` plugin.

Can you review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ import torch
 class ForceModule(torch.nn.Module):
     """A central harmonic potential as a static compute graph"""
     def forward(self, positions):
-        """The forward method returns the (1,1) energy tensor computed from the (nparticles,3) positions tensor.
+        """The forward method returns the energy computed from positions.
 
         Parameters
         ----------
@@ -132,7 +132,7 @@ particle position to translate all of them into a single periodic cell:
 class ForceModule(torch.nn.Module):
     """A central harmonic force with periodic boundary conditions"""
     def forward(self, positions, boxvectors):
-        """The forward method returns the (1,1) energy tensor computed from the (nparticles,3) positions tensor.
+        """The forward method returns the energy computed from positions.
 
         Parameters
         ----------
@@ -168,7 +168,7 @@ passed as an argument to `forward()`.
 class ForceModule(torch.nn.Module):
     """A central harmonic force with a user-defined global scale parameter"""
     def forward(self, positions, scale):
-        """The forward method returns the (1,1) energy tensor computed from the (nparticles,3) positions tensor.
+        """The forward method returns the energy computed from positions.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


jchodera:
It might be easiest to [view file](https://github.com/openmm/openmm-torch/blob/9ccde5ebc967eed2e0e43b39cbaa22400f8ac3cd/README.md) rather than just look at the diff.

jchodera:
@peastman : Just pinging again on this.

peastman:
I'll take a look when I can. We had a huge wind storm on Monday night, and
the power has been off ever since. Still no word on when they'll have it
back on. So I may be mostly out of touch for a bit.

jchodera:
Oh no! Hope there are at least no fires, and the power comes back on soon!

README.md
```diff
@@ -1,137 +1,196 @@
...
-=====
+Using the OpenMM PyTorch plugin
+===============================
+
+Exporting a PyTorch model for use in OpenMM
+-------------------------------------------
+
+The first step is to create a PyTorch model defining the calculation to perform.  
+It should take particle positions in nanometers (in the form of a `torch.Tensor` of shape `(nparticles,3)` and precision `torch.float64` as its input,
+and return the potential energy in kJ/mol as a `torch.Tensor` of shape `(1,1)` and precision `torch.float64` as output.
```
peastman, line 72: PyTorch models don't need to specify the precision of their inputs.  They handle whatever is given to them.  In this case, they might be either single or double precision, depending on the mode specified when creating the OpenMM Context.  The precision of the output also isn't important.  It will get cast as necessary.

The output should be a scalar, not a 2D tensor.

jchodera: @peastman : So the entire PyTorch model is in a single common precision? What if you specify arguments with different precisions?

jchodera: @peastman : I've addressed this comment.

jchodera:
@peastman: Any other changes? I'd love to tweet about this once we merge in the updated README.

README.md
peastman: The model can do whatever it wants with dtypes.  It will receive as input a tensor that could be either single or double precision.  If it just does calculations on it, the output will end up with the same type.

```python
>>> a = torch.tensor([1,2,3], dtype=torch.float32)
>>> torch.sum(a).dtype
torch.float32
```

But if you want to ensure calculations get done in a particular precision, you can include whatever casts you want.

```python
>>> b = a.to(torch.float64)
>>> torch.sum(b).dtype
torch.float64
```

So your model can do whatever it wants.  It just needs to ultimately return the energy, and we don't care what precision the return value has.

peastman: The docstrings still say, "The forward method returns the (1,1) energy tensor..."

jchodera: Fixed!

PR #25, opened 2021-01-28
add CI

### LINKED ISSUE #22: Set up continuous integration
@jaimergp : Any ideas on how we might be able to set up CI here now that we are able to install `openmm` an `pytorch` from conda-forge? 

In principle, we should at least be able to test the Reference and OpenCL platforms, and we may even be able to test the CUDA platform if we wanted to wire in @swails Jenkins until GPU CI becomes available.

jaimergp:
Actual checks running on my fork https://github.com/jaimergp/openmm-torch/pull/1. They won't show up until this is merged, I guess.

Follows same approach as https://github.com/openmm/openmm/pull/2964, but stripped of some features that are not as relevant here:

* Only Linux and OSX
* No multiarch
* No ccache
* No AMD OpenCL (to be discussed)

To do:
 
* Discuss build matrix

--- 

Closes #22


jaimergp:
@peastman This is ready for review. Check actual logs at https://github.com/jaimergp/openmm-torch/pull/1 (they won't show here until this is merged).

peastman:
That looks great!  I guess there's no way we can run the CUDA tests.  Would it make sense to install a CPU based OpenCL on the Linux build so it can run the OpenCL tests?

jaimergp:
Oh, right, I'll see if pocl or oclgrind help here. And yes, for CUDA tests we need a GPU via Jenkins or similar.

jaimergp:
Ok, took some debugging but we are good to go now!

peastman:
Excellent!  Shall I merge this?

jaimergp:
If you are happy with the current matrix, yes! There's only one thing I don't understand: [these missing PyTorch libraries](https://github.com/jaimergp/openmm-torch/runs/1810296238#step:10:23). The tests do pass, so I am puzzled...

jaimergp:
I guess it has to do with the fact that PyTorch is packaged with all the contents under `$PREFIX/lib/pythonX.Y/site-packages/torch`. So the libs are at `$PREFIX/lib/pythonX.Y/site-packages/torch/lib` instead of `$PREFIX/lib`... 🤷

peastman:
Do you need to add another directory to `LD_LIBRARY_PATH`?

jaimergp:
@peastman Yep, that did it! Thanks! I also added the Python tests (I didn't see they were present in this repo). I think this is ready now!

PR #46, opened 2021-09-27
Fix the force accumulation in the reference platform

raimis:
The forces should be accumulated rather than overwritten.


peastman:
Thanks!

PR #52, opened 2022-01-03
Models can directly output forces

### LINKED ISSUE #48: Allow the ML model to compute forces explicitly
Currently the atomic forces are computed implicitly by differentiating the potential energy returned by the ML model.

We should allow the ML model to compute force explicitly (which aren't necessary the derivatives of the potential energy). This would facilitate the implementation of some ML-based enhanced sampling algorithms (https://github.com/openmm/openmm/issues/3268).

peastman:
Fixes #48.  This allows the model to optionally return both energy and forces, instead of relying on backpropagation for the forces.


peastman:
I'm not sure what's causing the CI failures.  The Linux build gets this error:

>  /usr/share/miniconda/envs/build/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/bin/ld: /usr/share/miniconda/envs/build/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so: undefined reference to `memcpy@GLIBC_2.14'

It looks like PyTorch is compiled against a very old version of glibc that isn't available on the build machine.  But we get it from conda-forge, so shouldn't it use the same version as everything else?

The Mac build fails with this error when building the Python package.

```
Installed /usr/local/miniconda/envs/build/lib/python3.9/site-packages/openmmtorch-1.0-py3.9-macosx-10.9-x86_64.egg
Processing dependencies for openmmtorch==1.0
Searching for openmmtorch==1.0
Reading https://pypi.org/simple/openmmtorch/
Couldn't find index page for 'openmmtorch' (maybe misspelled?)
Scanning index of all packages (this may take a while)
Reading https://pypi.org/simple/
No local packages or working download links found for openmmtorch==1.0
error: Could not find suitable distribution for Requirement.parse('openmmtorch==1.0')
```

I don't understand that.  It claims to have just installed openmmtorch 1.0, and then immediately says that it can't find openmmtorch 1.0.

Any idea what's causing either of these?

raimis:
I'll try to solved the CI issue with https://github.com/openmm/openmm-torch/pull/53

peastman:
CI is passing.  Any comments on this, or should I merge it?

raimis (approved):
This looks good.

peastman:
Thanks!

Issue #59, opened 2022-01-21
Add example of using NNPOps with openmm-torch?


jchodera:
@peastman: Can we add an example script of how to use NNPOps directly within openmm-torch?

cc @dominicrufa @jeherr @raimis

peastman:
We need to finish making NNPOps conda installable first.

jchodera:
@peastman: There's [now a test conda package](https://github.com/openmm/NNPOps/issues/26#issuecomment-1017772953), but nobody is sure how to test it without an example of how to use it.

raimis:
I'll make an example.

raimis:
@jchodera I have created a draft of a tutorial (#62). You can see it better here: https://github.com/raimis/openmm-torch/blob/example/tutorials/openmm-torch-nnpops.ipynb

Or just [![Open On Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/raimis/openmm-torch/blob/example/tutorials/openmm-torch-nnpops.ipynb)

raimis:
The tutorials is done (#62). @jchodera @peastman could you check it?

raimis:
@jchodera could tweet about the new tutorial: https://github.com/openmm/openmm-torch/blob/master/tutorials/openmm-torch-nnpops.ipynb Also, this should be considered beta, but we need people trying it and find the remaining issues.

raimis:
I keep this issue open as a reminder to update the tutorial, when following issue are fixed:

- [ ]  An official NNPOps package is still not available: https://github.com/openmm/NNPOps/issues/26
- [ ] `Context.setVelocitiesToTemperature` is broken (https://github.com/openmm/openmm-torch/issues/61) and we need an OpenMM release with https://github.com/openmm/openmm/pull/3424.

jchodera:
This is waiting for the beta releases.

Issue #61, opened 2022-01-25
Setting velocities trigger an exception


raimis:
I tried to worn on #59 and got into an issue.

This is a simplified script: 
```python
import numpy as np
import torch as pt
from openmm import Context, Platform, System, VerletIntegrator
from openmmtorch import TorchForce

class NNP(pt.nn.Module):
  def __init__(self):
    super().__init__()

  def forward(self, positions):
    return pt.sum(positions)

pt.jit.script(NNP()).save('model.pt')

num_atoms = 3
system = System()
for _ in range(num_atoms):
    system.addParticle(1.0)
system.addForce(TorchForce('model.pt'))

integrator = VerletIntegrator(1.0)

context = Context(system, integrator, Platform.getPlatformByName('CUDA'))
context.setPositions(np.random.rand(num_atoms, 3))
context.setVelocitiesToTemperature(300)

integrator.step(1000)
```
which triggers an exception: 
```
Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.
Traceback (most recent call last):
  File "test.py", line 25, in <module>
    context.setVelocitiesToTemperature(300)
  File "/shared2/raimis/opt/miniconda/envs/ot_tut/lib/python3.7/site-packages/openmm/openmm.py", line 16068, in setVelocitiesToTemperature
    return _openmm.Context_setVelocitiesToTemperature(self, *args)
openmm.OpenMMException: The autograd engine was called while holding the GIL. If you are using the C++ API, the autograd engine is an expensive operation that does not require the GIL to be held so you should release it with 'pybind11::gil_scoped_release no_gil;'. If you are not using the C++ API, please report a bug to the pytorch team.
Exception raised from execute at /home/conda/feedstock_root/build_artifacts/pytorch-recipe_1635005512693/work/torch/csrc/autograd/python_engine.cpp:123 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x6a (0x7f1114a3198a in /shared2/raimis/opt/miniconda/envs/ot_tut/lib/python3.7/site-packages/torch/lib/libc10.so)
frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, char const*) + 0xd4 (0x7f1114a2d494 in /shared2/raimis/opt/miniconda/envs/ot_tut/lib/python3.7/site-packages/torch/lib/libc10.so)
[...]

PR #62, opened 2022-02-02
The first tutorial of OpenMM-Torch with NNPOps

raimis:
- [x] Write a tutorial
- [x] Add an automated test
- [x] Add a link to `README.md`

Solves #59

Try on Colab: https://colab.research.google.com/github/raimis/openmm-torch/blob/example/tutorials/openmm-torch-nnpops.ipynb


raimis:
The first tutorial of OpenMM-Torch is done.

There are two outstanding issue:
- An official NNPOps package is still not available: https://github.com/openmm/NNPOps/issues/26
- `Context.setVelocitiesToTemperature` is broken (https://github.com/openmm/openmm-torch/issues/61) and we need an OpenMM release with https://github.com/openmm/openmm/pull/3424.

Anyway, an imperfect tutorial is better than no tutorial.

peastman (approved):
This is really nice!

tutorials/openmm-torch-nnpops.ipynb
```diff
@@ -0,0 +1,440 @@
...
+        "# Create an integrator with a time step of 1 fs\n",
+        "temperature = 298.15 * kelvin\n",
+        "frictionCoeff = 1 / picosecond\n",
+        "timeStep = 1 * femtosecond\n",
+        "integrator = LangevinMiddleIntegrator(temperature, frictionCoeff, timeStep)\n",
+        "\n",
+        "# Create a simulation and set the initial positions and velocities\n",
+        "simulation = Simulation(ala2.topology, ala2.system, integrator)\n",
+        "simulation.context.setPositions(ala2.positions)\n",
+        "# simulation.context.setVelocitiesToTemperature(temperature) # This does not work (https://github.com/openmm/openmm-torch/issues/61)\n",
```
mikemhenry, line 326: Does this work now that https://github.com/openmm/openmm-torch/issues/61 is closed?

jchodera:
Would be great to get feedback from @dominicrufa too!

tutorials/openmm-torch-nnpops.ipynb
```diff
@@ -0,0 +1,440 @@
...
+            "🩹 Patching environment...\n",
+            "⏲ Done in 0:00:21\n",
+            "🔁 Restarting kernel...\n"
+          ]
+        }
+      ],
+      "source": [
+        "!pip install -q condacolab\n",
+        "import condacolab\n",
+        "condacolab.install_miniconda()"
```
jchodera, line 61: What about using just `condacolab.install()`, which will install `mambaforge` (miniforge + mamba) instead of `miniconda`? This can speed up installs.

tutorials/openmm-torch-nnpops.ipynb
```diff
@@ -0,0 +1,440 @@
...
+      "cell_type": "code",
+      "execution_count": 2,
+      "metadata": {
+        "id": "-qFE_cKF5958"
+      },
+      "outputs": [],
+      "source": [
+        "# Note: Remove \"mmh\" when NNPOps in available on conda-forge (https://github.com/openmm/NNPOps/issues/26)\n",
+        "# Note: \"cudatoollit=11.2\" is need to override a pinning in conda-colab\n",
+        "!conda install -q -c conda-forge -c mmh \\\n",
```
jchodera, line 87: If we use `mamba` instead of `conda` here, it can speed up installs significantly (presuming we also used mambaforge above).

@mikemhenry : Are we ready to drop the `-c mmh` yet?

tutorials/openmm-torch-nnpops.ipynb
```diff
@@ -0,0 +1,440 @@
...
+          ]
+        }
+      ],
+      "source": [
+        "import openmmtools\n",
+        "\n",
+        "# Get the system of alanine dipeptide\n",
+        "ala2 = openmmtools.testsystems.AlanineDipeptideVacuum(constraints=None)\n",
+        "\n",
+        "# Remove MM forces\n",
```
jchodera, line 129: What about refactoring the process of removing MM forces and creating/adding the ML potential into a small function folks can copy-and-paste into their own code?

For example, this code could be bundled together:
```python
def replace_mm_with_nnp(system, topology, NNP):
    """
    Create an instance of the specified machine learning potential and replace all MM forces with the ML potential.

    """
    # Remove MM forces
    while system.getNumForces() > 0:
        system.removeForce(0)

    # Assert no forces or constraints remain
    assert system.getNumConstraints() == 0 
    assert system.getNumForces() == 0

    # Get atomic numbers
    atomic_numbers = [ atom.element.atomic_number for atom in topology.atoms() ]

    # Create an instance of the model
    nnp = NNP(atomic_numbers)
   
    # Save the NNP to a file and load it with OpenMM-Torch
    # TODO: Use a temporary file instead
    pt.jit.script(nnp).save('model.pt')
    force = TorchForce('model.pt')

    # Add the NNP to the system
    system.addForce(force)
    assert ala2.system.getNumForces() == 1 
```

tutorials/openmm-torch-nnpops.ipynb
```diff
@@ -0,0 +1,440 @@
...
+        "from NNPOps import OptimizedTorchANI\n",
+        "\n",
+        "class NNP(pt.nn.Module):\n",
+        "\n",
+        "  def __init__(self, atomic_numbers):\n",
+        "\n",
+        "    super().__init__()\n",
+        "\n",
+        "    # Store the atomic numbers\n",
+        "    self.atomic_numbers = pt.tensor(atomic_numbers).unsqueeze(0)\n",
```
jchodera, line 200: How does this work for atoms that should _not_ have their potential computed by the NNP? Would we use atomic numbers of `0` for those atoms? For example, if we wanted to only compute ligand energies for a solvated protein:ligand system?

jchodera (commented):
This is really great! Just had some minor comments.

tutorials/openmm-torch-nnpops.ipynb
peastman: I think that's beyond the scope of this first tutorial.  Mixing ML with MM could be a more advanced tutorial.  Or we could just refer people to OpenMM-ML, which has code to automate it.

jchodera: It seems weird not to even mention how it would work, since this is a major use case. If folks wanted to just simulate only ML, why would they use OpenMM? 

I agree a separate tutorial would be useful!

What about periodic ML systems? Do we support those? Or just non-periodic systems?

Maybe we could have three tutorials:
* Vacuum ML
* Periodic ML
* Hybrid ML/MM periodic

peastman: Periodic systems just take a minor change to the model.  See https://github.com/openmm/openmm-torch/blob/master/README.md.

tutorials/openmm-torch-nnpops.ipynb
mikemhenry: Not yet, if we can't drop it by the time we want to merge the tutorial, we can make an issue to drop it so I won't forget once we get NNPOps on conda-forge

tutorials/openmm-torch-nnpops.ipynb
raimis: I tried `mamba`. It works on a local machine, but fail on Colab due to a pinning issue.

tutorials/openmm-torch-nnpops.ipynb
raimis: As commented below, this a low-level tutorial. All the convenience functionality should be in OpenMM-ML.

raimis:
No more comments, so I merge the tutorial. We could revise it when there will be more tutorials in place and we will have a better idea.

jchodera:
@dominicrufa: Can you provide some feedback too?

mikemhenry:
@raimis nnpops is now on conda-forge :tada:  `mamba info nnpops` So you can drop the `-c -mmh` now

PR #68, opened 2022-02-24
Support CUDA Graphs

raimis:
Inspired by the significant performance increase with CUDA Graphs (https://github.com/torchmd/torchmd-net/issues/48#issuecomment-1048899019), this PR implemented the support of CUDA Graphs to accelerate PyTorch models (https://pytorch.org/blog/accelerating-pytorch-with-cuda-graphs/).

- [x] Separate the graphable operations
- [x] Make the input and output tensors persistent
- [x] Capture the graph
- [x] Expose an option to enable the graph
- [ ] Serialize the options
- [x] Conditionally build for the old PyTorch version without the support of CUDA Graphs
- [x] Add tests
- [ ] Update documentation


raimis:
@peastman I need your opinion.

We need to extend the API for platform-specific options, e.g. the usage of CUDA Graphs.

My proposal is to add a simple dictionary-style API: 
```python
TorchForce.getPropertyValue(name: str) -> str
TorchForce.setPropertyValue(name: str, value: str)
```
For convenience, the constructor can accept a dictionary directly: 
```python
TorchForce(str, Optional[Dict[str, str]])
```
The names and values are checked when a platform-specific `CalcTorchForceKernel` is initialized.

peastman:
That sounds reasonable.

PR #70, opened 2022-02-28
Move the PyTorch module to a correct device

raimis:
On a multi-GPU system, the Pytorch module should be moved to a selected device, but it was always moved to `cuda:0`.

- [x] Move the module to a correct device
- [x] Add a test


raimis:
@peastman could you review?

peastman (approved):
Looks good.

PR #73, opened 2022-04-08
Update the test models and CI dependencies

raimis:
- [x] Update dependencies
- [x] Update to PyTorch 1.11
  - [x] Regenerate the test models


raimis:
Uff! Still the dependencies are missing for Python 3.10

raimis:
So, at the moment, it is only possible to update CUDA.

raimis:
- Fixed the tests with PyTorch 1.11. The test models have been generated with some old PyTorch and PyTorch 1.11 broke backward compatibility. This should allow to build the `conda-forge` packages  for PyTorch 1.11 (https://github.com/conda-forge/openmm-torch-feedstock/pull/17).
- After, it will be possible to fix the tutorial test, which are broken because the `conda-forge` migrated to PyTorch 1.11 (https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2812) and OpenMM-Torch was left behind.

tests/generate.py
```diff
@@ -0,0 +1,24 @@
+import torch as pt
+
+class Central(pt.nn.Module):
+    def forward(self, pos):
+        return pos.pow(2).sum(dim=1).abs().sum()
```
peastman, line 5: I think this and the similar ones below can be simplified to just `pos.pow(2).sum()`.  Squaring the elements guarantees all of them are positive, so `abs()` does nothing.  And that means the intermediate sum also is unnecessary

raimis: Good point.

raimis: Done!

raimis:
@peastman any more comments?

peastman:
Looks good!  I see there's a test failure.  It looks like the same problem as #76.

mikemhenry:
I'll see if there is a quick re-render and rebuild of nnpops fixes this.

PR #77, opened 2022-06-03
Fix the tutorial

raimis:
- [x] Explicitly install Miniconda
- [x] Update dependencies


raimis:
@peastman could you review?

peastman (approved):
I haven't tested it, but everything looks reasonable.  As long as it works, I don't have any changes to suggest.

PR #49, opened 2021-12-10
Fix the synchronisation between the CUDA contexts

raimis:
- [x] Synchronize the switching between CUDA contexts
- [x] Remove a redundant coping of forces

Related with #47

PR #80, opened 2022-06-15
Fix interoperability with CustomCVForce

raimis:
- [x] Fix interoperability with `OpenMM::CustomCVForce`
- [x] Tests

### DIFF
--- a/platforms/cuda/src/CudaTorchKernels.cpp
+++ b/platforms/cuda/src/CudaTorchKernels.cpp
@@ -97,6 +97,9 @@ void CudaCalcTorchForceKernel::initialize(const System& system, const TorchForce
 double CudaCalcTorchForceKernel::execute(ContextImpl& context, bool includeForces, bool includeEnergy) {
     int numParticles = cu.getNumAtoms();
 
+    // Push to the PyTorch context
+    CHECK_RESULT(cuCtxPushCurrent(primaryContext), "Failed to push the CUDA context");
+
     // Get pointers to the atomic positions and simulation box
     void* posData;
     void* boxData;
@@ -117,9 +120,6 @@ double CudaCalcTorchForceKernel::execute(ContextImpl& context, bool includeForce
         cu.executeKernel(copyInputsKernel, inputArgs, numParticles);
         CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context"); // Synchronize before switching to the PyTorch context
     }
-
-    // Push to the PyTorch context
-    CHECK_RESULT(cuCtxPushCurrent(primaryContext), "Failed to push the CUDA context");
 
     // Prepare the input of the PyTorch model
     vector<torch::jit::IValue> inputs = {posTensor};
--- a/platforms/cuda/src/CudaTorchKernels.cpp
+++ b/platforms/cuda/src/CudaTorchKernels.cpp
@@ -51,6 +51,7 @@ if (result != CUDA_SUCCESS) { \
 
 CudaCalcTorchForceKernel::CudaCalcTorchForceKernel(string name, const Platform& platform, CudaContext& cu) :
         CalcTorchForceKernel(name, platform), hasInitializedKernel(false), cu(cu) {
+    // Explicitly activate the primary context
     CHECK_RESULT(cuDevicePrimaryCtxRetain(&primaryContext, cu.getDevice()), "Failed to retain the primary context");
 }
 
@@ -66,10 +67,12 @@ void CudaCalcTorchForceKernel::initialize(const System& system, const TorchForce
         globalNames.push_back(force.getGlobalParameterName(i));
     int numParticles = system.getNumParticles();
 
-    // Switch to the PyToch context
-    CUcontext context;
-    CHECK_RESULT(cuCtxGetCurrent(&context), "Failed to get the CUDA context");
-    if (context)
+    // Switch to the PyTorch context
+    // NOTE: Pytorch is always using the primary context.
+    //       It makes the primary context current, if it is not a case.
+    CUcontext ctx;
+    CHECK_RESULT(cuCtxGetCurrent(&ctx), "Failed to get the CUDA context");
+    if (ctx)
         CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context"); // Synchronize before switching to the PyTorch context
     CHECK_RESULT(cuCtxPushCurrent(primaryContext), "Failed to push the CUDA context");
 
@@ -82,10 +85,10 @@ void CudaCalcTorchForceKernel::initialize(const System& system, const TorchForce
     posTensor = torch::empty({numParticles, 3}, options.requires_grad(!outputsForces));
     boxTensor = torch::empty({3, 3}, options);
 
-    // Switch to the CUDA context
-    CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context");
-    CHECK_RESULT(cuCtxPopCurrent(&context), "Failed to pop the CUDA context");
-    assert(primaryContext == context); // Sanity check that PyTorch haven't messed up the context
+    // Switch to the OpenMM context
+    CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context"); // Synchronize before switching to the OpenMM context
+    CHECK_RESULT(cuCtxPopCurrent(&ctx), "Failed to pop the CUDA context");
+    assert(primaryContext == ctx); // Check that PyTorch haven't messed up the context
     ContextSelector selector(cu); // Switch to the OpenMM context
 
     // Initialize CUDA objects for OpenMM-Torch
--- a/platforms/cuda/src/CudaTorchKernels.cpp
+++ b/platforms/cuda/src/CudaTorchKernels.cpp
@@ -67,7 +67,10 @@ void CudaCalcTorchForceKernel::initialize(const System& system, const TorchForce
     int numParticles = system.getNumParticles();
 
     // Switch to the PyToch context
-    // CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context"); // Synchronize before switching to the PyTorch context
+    CUcontext context;
+    CHECK_RESULT(cuCtxGetCurrent(&context), "Failed to get the CUDA context");
+    if (context)
+        CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context"); // Synchronize before switching to the PyTorch context
     CHECK_RESULT(cuCtxPushCurrent(primaryContext), "Failed to push the CUDA context");
 
     // Initialize CUDA objects for PyTorch
@@ -81,7 +84,6 @@ void CudaCalcTorchForceKernel::initialize(const System& system, const TorchForce
 
     // Switch to the CUDA context
     CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context");
-    CUcontext context;
     CHECK_RESULT(cuCtxPopCurrent(&context), "Failed to pop the CUDA context");
     assert(primaryContext == context); // Sanity check that PyTorch haven't messed up the context
     ContextSelector selector(cu); // Switch to the OpenMM context
--- a/platforms/cuda/src/CudaTorchKernels.cpp
+++ b/platforms/cuda/src/CudaTorchKernels.cpp
@@ -66,6 +66,10 @@ void CudaCalcTorchForceKernel::initialize(const System& system, const TorchForce
         globalNames.push_back(force.getGlobalParameterName(i));
     int numParticles = system.getNumParticles();
 
+    // Switch to the PyToch context
+    // CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context"); // Synchronize before switching to the PyTorch context
+    CHECK_RESULT(cuCtxPushCurrent(primaryContext), "Failed to push the CUDA context");
+
     // Initialize CUDA objects for PyTorch
     const torch::Device device(torch::kCUDA, cu.getDeviceIndex()); // This implicitly initialize PyTorch
     module.to(device);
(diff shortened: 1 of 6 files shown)


raimis:
I have just built a package for https://github.com/openmm/openmm-torch/pull/80/commits/7482623141d09f14dce506ac28cdf9b6017ad38b:

```bash
conda install -c raimis/label/debug -c conda-forge openmm-torch=0.8rc0
```

@dominicrufa could you check if it fixes https://github.com/openmm/openmm-ml/issues/32?

@Hong-Rui could you check if it fixes https://github.com/openmm/openmm/issues/3588?

platforms/cuda/src/CudaTorchKernels.cpp
```diff
@@ -49,7 +49,13 @@ if (result != CUDA_SUCCESS) { \
     throw OpenMMException(m.str());\
 }
 
+CudaCalcTorchForceKernel::CudaCalcTorchForceKernel(string name, const Platform& platform, CudaContext& cu) :
+        CalcTorchForceKernel(name, platform), hasInitializedKernel(false), cu(cu) {
+    CHECK_RESULT(cuDevicePrimaryCtxRetain(&primaryContext, cu.getDevice()), "Failed to retain the primary context");
```
peastman, line 54: I think this is incorrect in a couple of ways.  First, there's no guarantee that PyTorch will have set up the primary context yet at this point.  Inside `initialize()` we ensure it has been initialized with the line

https://github.com/openmm/openmm-torch/blob/661b004eb1a2b1c9cca8c8c857d3336f17762017/platforms/cuda/src/CudaTorchKernels.cpp#L64

Depending on what else has happened earlier, it might have already been initialized at this point but it's not guaranteed.

Second, by recording the primary context here you're assuming it will be created once and never changed.  That's the typical pattern with the runtime API, but it isn't guaranteed.  `cudaDeviceReset()` can be used to delete the primary context and create a new one.  We shouldn't make assumptions about what PyTorch might do.  The safe approach is to query the primary context immediately before pushing it with `cuCtxPushCurrent()`.

platforms/cuda/src/CudaTorchKernels.cpp
```diff
@@ -60,6 +66,13 @@ void CudaCalcTorchForceKernel::initialize(const System& system, const TorchForce
         globalNames.push_back(force.getGlobalParameterName(i));
     int numParticles = system.getNumParticles();
 
+    // Switch to the PyToch context
+    CUcontext context;
+    CHECK_RESULT(cuCtxGetCurrent(&context), "Failed to get the CUDA context");
+    if (context)
+        CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context"); // Synchronize before switching to the PyTorch context
+    CHECK_RESULT(cuCtxPushCurrent(primaryContext), "Failed to push the CUDA context");
+
```
peastman, line 75: Can you explain what this code is for?  This method doesn't involve any communication between the PyTorch and OpenMM contexts, so synchronization shouldn't be necessary.  And as noted above, the primary context might not have even been created yet at this point.

dominicrufa:
@raimis, i can confirm that the fix works the `nnpops`-equipped `TestMLPotential.py` test from `openmm-ml` and I can run simulations with it, as well. I can update the `TestMLPotential.py` test in `openmm-ml` and add the different test cases now. 

However, I don't know if there are any bugs that will creep up re: comments that @peastman  made above. will have to do some deeper digging to see if these are actually problematic.

platforms/cuda/src/CudaTorchKernels.cpp
raimis: Regarding the second point, `cudaDeviceReset` (https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gef69dd5c6d0206c2b8d099abac61f217) destroys all the contexts, not only the primary one. Also, all the memory allocations are wiped out. So, it is game over, the only thing OpenMM-Torch can do is to crash.

raimis: Regarding the first point, the primary context will be activate in the constructor by `cuDevicePrimaryCtxRetain` (if not yet active) (https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__PRIMARY__CTX.html#group__CUDA__PRIMARY__CTX_1g9051f2d5c31501997a6cb0530290a300) and it is guaranteed to be active until `cuDevicePrimaryCtxRelease` is called in the destructor.

platforms/cuda/src/CudaTorchKernels.cpp
raimis: This ensures that `initialize` from a previous forces hasn't left some kernel running the OpenMM context. As, we seen before, bad thing happen if the context switches aren't synchronized. And as noted above, the primary context is activated in the constructor.

platforms/cuda/src/CudaTorchKernels.cpp
peastman: > cudaDeviceReset destroys all the contexts, not only the primary one.

I don't think that's correct.  The documentation you linked refers to the runtime API, which treats "device" and "context" as synonymous.  Calling methods on the runtime API doesn't affect resources allocated separately by the lower level driver API.

platforms/cuda/src/CudaTorchKernels.cpp
peastman: There is no possibility of that happening.  This method is called immediately after the CudaContext is created, before `initialize()` has even been called yet on it.  We have a new, just created CUcontext.  Even if there were somehow a kernel running, it still wouldn't matter.  Nothing in this method does any communication between the PyTorch and OpenMM contexts.

platforms/cuda/src/CudaTorchKernels.cpp
peastman: This documentation in the [CUDA Programming Guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization) is a bit more explicit about it:

> When a host thread calls cudaDeviceReset(), this destroys the primary context of the device the host thread currently operates on (i.e., the current device as defined in [Device Selection](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-selection)). The next runtime function call made by any host thread that has this device as current will create a new primary context for this device.

raimis:
I have just rebuilt the package for https://github.com/openmm/openmm-torch/pull/80/commits/75e30361fd36c3503dbac4b13b4df7014da52ca3:

```
conda install -c raimis/label/debug -c conda-forge openmm-torch=0.8rc1
```

platforms/cuda/src/CudaTorchKernels.cpp
peastman: Correct.

peastman:
Let's switch back to my fix in #78.  It was a much simpler change, and we already verified it fixed the problem.  The changes currently in this PR add unnecessary complexity, create new opportunities for bugs, and hurt performance through unnecessary calls to `cuCtxSynchronize()`, which is a very expensive function.

raimis:
If the contexts are not synchronized, there may be memory corruptions as before #49.

peastman:
The only time you need synchronization is when you're copying information between the two contexts.  The rest of the time, they each run independently and don't interfere with each other.

Hong-Rui:
Hi, regarding your discussions above, I'm just curious that if we messed up with the cuda contexts or make things corrupts, does it affect the OpenMM's force calculations such that it might occasionally produce wrong or random results when doing integrations?

I'm asking this because I'm afraid that if sometimes the NN biased simulations would just run silently even if some wrong cuda context operations happens inside. Is this possible? or does it always report some cuda error like we have met? Thanks.

Hong-Rui:
Hi, regarding your discussions above, I'm just curious that if we messed up with the cuda contexts or make things corrupts, does it affect the OpenMM's force calculations such that it might occasionally produce wrong or random results when doing integrations?

I'm asking this because I'm afraid that sometimes the NN biased simulations would just run silently even if it contains some wrong cuda context operations that would mess things up. Is this possible? or do we always get some cuda error reports like we have met when this happens? Thanks.

Hong-Rui:
Thanks, Peter.
I noticed that there will be significant performance slowdown when using torch force with OpenMM. The slowdown arises from extra calculations performed at each MD step, such as NN model inference, some other forcefield-related calculations inside the torch force. I think it's better that we fix the cuda context management bug in a way that won't introduce too many performances slow down.

peastman:
We need to finish this up.  It's blocking the OpenMM 8.0 beta.  If you want, I can push the changes.

Hong-Rui:
Yeah, but I'm not familiar with cuda context managements so cannot comment on what you discussed above. And it seems that you haven't meet agrrements with each other on how to deal with this correctly...
For users like me, I thought we just want a fix that won't cause further performance slown down and won't introduce other bugs or possible memory corruptions...?

Hong-Rui:
I'm currently using the OpenMM master (compiled by myself) and the openmm-torch built by @raimis above to perform my simulations as a temperal production environment.

raimis:
@Hong-Rui 

> Hi, regarding your discussions above, I'm just curious that if we messed up with the cuda contexts or make things corrupts, does it affect the OpenMM's force calculations such that it might occasionally produce wrong or random results when doing integrations?

Yes, it can produce incorrect results and not crash immediately. 

Before #49, the situation was that simulations of small molecule (e.g. aspirin) were running fine. Later, when I stared with larger molecules (https://arxiv.org/abs/2201.08110), simulations were crashing randomly. Some ran just a few steps, some continued for a few ns. The problem was that due to missing synchronization, forces at some steps were incorrect and in many case the magnitudes of incorrect force within plausible limits. So, the issue wasn't obvious and took a few weeks to sort out.

Another complication was that the simulations on a GPU with an attached screen was more stable, than on a dedicated GPU of the same model. I guess, the competing graphics kernels were making OpenMM-Torch kernels to run in more serial way. So, it was hiding the issue.

raimis:
For this reason, I'm a bit caution about the CUDA contexts. It is a low-level API and give all the opportunities to shoot yourself in the foot.

I can take out some `cuCtxSynchronize()` as @peastman insists, but @dominicrufa and @Hong-Rui will have to test with some non-trivial systems.

platforms/cuda/src/CudaTorchKernels.cpp
raimis: Even if it destroy just the primary context, it is still gave over. PyTorch stores the NNP model on the primary context and it will be gone.

platforms/cuda/src/CudaTorchKernels.cpp
raimis: Removed that check (https://github.com/openmm/openmm-torch/pull/80/commits/d3d060b36f6b1bad07154f626f1fefe0e255fc0b)

raimis:
I have just rebuilt the package for https://github.com/openmm/openmm-torch/pull/80/commits/d3d060b36f6b1bad07154f626f1fefe0e255fc0b:

```
conda install -c raimis/label/debug -c conda-forge openmm-torch=0.8rc2
```

platforms/cuda/src/CudaTorchKernels.cpp
```diff
@@ -94,13 +111,16 @@ double CudaCalcTorchForceKernel::execute(ContextImpl& context, bool includeForce
...
+        ContextSelector selector(cu); // Switch to the OpenMM context
         void* inputArgs[] = {&posData, &boxData, &cu.getPosq().getDevicePointer(), &cu.getAtomIndexArray().getDevicePointer(),
                 &numParticles, cu.getPeriodicBoxVecXPointer(), cu.getPeriodicBoxVecYPointer(), cu.getPeriodicBoxVecZPointer()};
         cu.executeKernel(copyInputsKernel, inputArgs, numParticles);
-        CHECK_RESULT(cuCtxSynchronize(), "Error synchronizing CUDA context"); // Synchronize before switching to the PyTorch context
+        CHECK_RESULT(cuCtxSynchronize(), "Failed to synchronize the CUDA context"); // Synchronize before switching to the PyTorch context
     }
 
+    // Push to the PyTorch context
+    CHECK_RESULT(cuCtxPushCurrent(primaryContext), "Failed to push the CUDA context");
```
peastman, line 122: If we want to be completely safe, we probably should move this up to the top of `execute()`.  Most likely it's ok to call `data_ptr()` on a tensor without setting the context first, but it seems good to be safe.

dominicrufa:
> For this reason, I'm a bit caution about the CUDA contexts. It is a low-level API and give all the opportunities to shoot yourself in the foot.
> 
> I can take out some `cuCtxSynchronize()` as @peastman insists, but @dominicrufa and @Hong-Rui will have to test with some non-trivial systems.

@raimis , I'm not sure i am following this thread completely, but I'd be happy to write a test or two for this seeing as I need to run simulations with this on some nontrivial systems anyway. by "nontrivial", presumably you mean treat a system of >50 atoms with the nnp? and make sure that the total energy/forces are consistent with nnp energies/forces not computed in the omm framework?

raimis:
@dominicrufa it would great if you can contribute more tests.

I meant with "non-trivial":
- Large systems and different NNP models
- System containing other forces (not just `TorchForce`) to avoid unforeseen interactions like this issue.

I guess, these integration/correctness test should be at OpenMM-ML level. So, they are as close as possible to production cases.

dominicrufa:
> @dominicrufa it would great if you can contribute more tests.
> 
> I meant with "non-trivial":
> 
> * Large systems and different NNP models
> * System containing other forces (not just `TorchForce`) to avoid unforeseen interactions like this issue.
> 
> I guess, these integration/correctness test should be at OpenMM-ML level. So, they are as close as possible to production cases.

@raimis , we've already refined a test in `openmm-ml` [here](https://github.com/openmm/openmm-ml/blob/main/test/TestMLPotential.py) that makes some assertions on a solvated alanine dipeptide with ani2x with/without nnpops (this is a mixed system with other, non `TorchForce` forces). I can extend this to run on ani1 and possibly make sure that the ani energy computed in openmm matches that of the standalone torch module (though might not be necessary?). 

when you refer to larger systems, do you mean more atoms treated with nnp? i can drop a larger solvated small molecule or peptide into the test. 

at the moment i am just struggling to understand the exact failure mode that you are concerned about.

platforms/cuda/src/CudaTorchKernels.cpp
peastman: If you can fix this last detail, we can get this merged and release the OpenMM 8 beta.

raimis: Done https://github.com/openmm/openmm-torch/pull/80/commits/beacb07aa6607a12d980920d7689a5c9b960c686!

raimis:
@peastman are you still reviewing or you just forget to approve?

raimis:
OK! I'm making a new release of OpenMM-Torch and push the new packages for OpenMM 7.7

PR #85, opened 2022-10-25
Update the tutorial for OpenMM 8.0beta

raimis:
- [x] Install MambaForge explicitly

### DIFF
--- a/tutorials/openmm-torch-nnpops.ipynb
+++ b/tutorials/openmm-torch-nnpops.ipynb
@@ -39,18 +39,18 @@
           "base_uri": "https://localhost:8080/"
         },
         "id": "EcB2lc1s3ZXG",
-        "outputId": "1ce0bad8-a050-48d5-c18a-9e49fbeb11c2"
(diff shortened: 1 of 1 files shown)


tutorials/openmm-torch-nnpops.ipynb
```diff
@@ -82,11 +82,17 @@
...
+        "!mamba install -q \\\n",
+        "               -c conda-forge/label/openmm_rc \\\n",
+        "               -c conda-forge/label/openmm-torch_rc \\\n",
+        "               -c conda-forge \\\n",
+        "               openmm==8.0.0beta \\\n",
+        "               openmm-torch==1.0beta \\\n",
+        "               nnpops==0.2 \\\n",
+        "               torchani==2.2.2 \\\n",
+        "               openmmtools \\\n",
+        "               &> /dev/null # Comment this line to see a log"
```
RaulPPelaez, line 95: This should be good to go with a some changes here, right?
I guess removing the channels and the betas should do the trick, also nnpops to 0.5?

PR #86, opened 2022-10-25
Pin swig to <4.1

raimis:
- [x] Pin `swig` to <4.1

`swig` 4.1.0 fails with the following error:
```
/usr/share/miniconda3/envs/build/include/swig/OpenMMSwigHeaders.i:12: Error: Unknown directive '%factory'.
```

### DIFF
--- a/devtools/conda-envs/build-macos-11.yml
+++ b/devtools/conda-envs/build-macos-11.yml
@@ -12,4 +12,4 @@ dependencies:
   - pytest
   - python
   - pytorch-cpu @PYTORCH_VERSION@
-  - swig
+  - swig <4.1
--- a/devtools/conda-envs/build-ubuntu-18.04.yml
+++ b/devtools/conda-envs/build-ubuntu-18.04.yml
@@ -15,6 +15,6 @@ dependencies:
   - pytest
   - python
   - pytorch-gpu @PYTORCH_VERSION@
-  - swig
+  - swig <4.1
   - sysroot_linux-64 2.17
   - torchani


raimis:
`swig` 4.1.0 was released today (https://anaconda.org/conda-forge/swig/files?version=4.1.0)

raimis:
@peastman could you review?

raimis:
@mikemhenry we may need to pin the feedstock or fix that header in OpenMM.

mikemhenry:
Builds here
https://github.com/conda-forge/openmm-torch-feedstock/pull/29
https://github.com/conda-forge/openmm-torch-feedstock/pull/28

peastman (approved):
Looks good.

peastman:
I can't reproduce this problem.  When I build both OpenMM and this plugin with swig 4.1.0, everything builds without problem.  Is there something else I need to do to reproduce it?

scal444:
I reproduced the problem and lost a few hours debugging before I went in to the devtools files to see what the working build requirements were. Peter I'm not sure why it doesn't trigger for you, but in the meantime it would be nice to add a check for the version, or put a warning in the documentation. Particularly since swig 4 shows up (for me) with the suggested [conda environment for openmm](http://docs.openmm.org/latest/userguide/library/02_compiling.html#other-required-software)

This was also reproducible for me on the [example plugin repo](https://github.com/openmm/openmmexampleplugin)

peastman:
I figured out why I couldn't reproduce it.  CMake was picking up /usr/bin/swig instead of the one from the conda environment.  I'll see if I can figure out the fix.

peastman:
See #96.  I think we were just missing a `%include`.

Issue #64, opened 2022-02-08
Improve the installation instruction from source


raimis:
The installation instructions from source (https://github.com/openmm/openmm-torch#building-from-source) should be improved. We already have two issue (#40, #63).

It would be useful to provide a Conda file for the build environment, like we have for NNPOps (https://github.com/openmm/NNPOps#from-the-source).

PR #89, opened 2022-11-11
Improve build from source instructions 

sef43:
Addressess #64 

* Provides a conda environment file
* update build instructions in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,43 +27,94 @@ If you don't have `conda` available, we recommend installing [Miniconda for Pyth
 Building from source
 --------------------
 
-This plugin uses [CMake](https://cmake.org/) as its build system.  
-Before compiling you must install [LibTorch](https://pytorch.org/cppdocs/installing.html), which is the PyTorch C++ API, by following the instructions at https://pytorch.org.
-You can then follow these steps:
+### Prerequisites
 
-1. Create a directory in which to build the plugin.
+- Minconda https://docs.conda.io/en/latest/miniconda.html#linux-installers
+- CUDA Toolkit https://developer.nvidia.com/cuda-downloads 
+- LibTorch which is the PyTorch C++ API, by following the instructions at https://pytorch.org/cppdocs/installing.html and https://pytorch.org. 
+    - to get a version for CUDA 1.16 used in our build example we use the commands:
+    ```
+        wget https://download.pytorch.org/libtorch/cu116/libtorch-cxx11-abi-shared-with-deps-1.13.0%2Bcu116.zip
+        unzip libtorch-cxx11-abi-shared-with-deps-1.13.0%2Bcu116.zip
+    ```
 
-2. Run the CMake GUI or `ccmake`, specifying your new directory as the build directory and the top
-level directory of this project as the source directory.
+### Build & install
 
-3. Press "Configure".  (Do not worry if it produces an error message about not being able to find PyTorch.)
+1. Get the source code
 
-4. Set `OPENMM_DIR` to point to the directory where OpenMM is installed.  This is needed to locate
-the OpenMM header files and libraries.  If you are unsure of what directory this is, the following
-script will print it out.
+   ```
+   git clone https://github.com/openmm/openmm-torch.git
+   cd openmm-torch
+   ```
 
-```python
-from simtk import openmm
-import os
-print(os.path.dirname(openmm.version.openmm_library_path))
+2. Set CUDA_HOME  (you may have a different path and version)
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -27,43 +27,94 @@ If you don't have `conda` available, we recommend installing [Miniconda for Pyth
...
-This plugin uses [CMake](https://cmake.org/) as its build system.  
-Before compiling you must install [LibTorch](https://pytorch.org/cppdocs/installing.html), which is the PyTorch C++ API, by following the instructions at https://pytorch.org.
-You can then follow these steps:
+### Prerequisites
 
-1. Create a directory in which to build the plugin.
+- Minconda https://docs.conda.io/en/latest/miniconda.html#linux-installers
+- CUDA Toolkit https://developer.nvidia.com/cuda-downloads 
+- LibTorch which is the PyTorch C++ API, by following the instructions at https://pytorch.org/cppdocs/installing.html and https://pytorch.org. 
+    - to get a version for CUDA 1.16 used in our build example we use the commands:
```
raimis, line 35: Please, don't use stuff from pytorch.org. Their binaries are incompatible with `conda-forge`.

environment.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
```
raimis, line 1: Just reuse the one from CI: https://github.com/openmm/openmm-torch/blob/master/devtools/conda-envs/build-ubuntu-18.04.yml

README.md
```diff
@@ -27,43 +27,94 @@ If you don't have `conda` available, we recommend installing [Miniconda for Pyth
...
+   ```
+   conda env create -n openmm-torch -f environment.yaml
+   conda activate openmm-torch
+   ```
+
+4. Configure 
+   ```
+   mkdir build && cd build
+
+   cmake .. -DPYTORCH_DIR=<path/to/libtorch> \
```
raimis, line 68: If you install the environment with `conda`, you already know what the paths are.

README.md
```diff
@@ -27,43 +27,94 @@ If you don't have `conda` available, we recommend installing [Miniconda for Pyth
...
 --------------------
 
-This plugin uses [CMake](https://cmake.org/) as its build system.  
-Before compiling you must install [LibTorch](https://pytorch.org/cppdocs/installing.html), which is the PyTorch C++ API, by following the instructions at https://pytorch.org.
-You can then follow these steps:
+### Prerequisites
 
-1. Create a directory in which to build the plugin.
+- Minconda https://docs.conda.io/en/latest/miniconda.html#linux-installers
+- CUDA Toolkit https://developer.nvidia.com/cuda-downloads
```
peastman, line 33: CUDA is only needed if you have a NVIDIA GPU you want to use with it.

README.md
peastman: We probably want to reconsider the build process and possibly make corresponding changes to the CMake script.  If you set `TORCH_DIR` (*not* `PYTORCH_DIR`) to the directory containing `TorchConfig.cmake`, it should be able to automatically find everything it needs.

peastman:
These instructions seem to be Linux specific.  We also support Mac and Windows.  Note that conda-forge does not currently provide PyTorch builds for WIndows, so in that case it's necessary to get it from the PyTorch website or the `pytorch` conda channel.

sef43:
I have made specific instructions for Linux (with and without CUDA) and MacOS.
One issue is that I think the conda-forge NNPOps only has linux with cuda versions available

raimis:
We should put the from-source installation instructions to a separate file (e.g. `INSTALL.md`) and link it from the main `READM.md`. Most users can install with `conda`, so they don't need to be distracted.

peastman:
>  One issue is that I think the conda-forge NNPOps only has linux with cuda versions available

That's because conda-forge doesn't provide PyTorch builds for Windows.  That makes building from source especially important, since it's the only option for Windows.

sef43:
For instructions for a building on Windows we will need to update `PythonInstall` and `python/setup.py` etc to work with windows. I am not familiar enough with the build system/Windows to do this.

I have been able to compile the c++ code on Windows (CPU only) using visual studio and this environment:
```
name: build
channels:
  - conda-forge
  - pytorch
dependencies:
  - openmm >=7.7
  - pip
  - pytest
  - python
  - pytorch::pytorch
  - pytorch::cpuonly 
  - swig <4.1
```
One quirk is that `pytorch` requires MKL, but the conda mkl for me did not seem to have all the files needed to compile so I had to download the full MKL installer from intel.

Issue #66, opened 2022-02-11
Simplify the creation of `TorchForce`


raimis:
In general, the creation of `TorchForce` goes like this:
```python
import torch as pt
from openmmtorch import TorchForce

# 1. create a PyTorch module
class Module(pt.nn.Module):
    def __init__(self):
        super().__init__();
    def forward(self, positions):
        return pt.sum(positions)

# 2. save the module to a file
module = Module()
pt.jit.script(module).save('module.pt')

# 3. create a force from the file
force = TorchForce('module.pt')
```

We could simplify the last two steps by allowing to pass an instance of `Module` directly to `TorchForce`:
```python
force = TorchForce(module)
```

For this, we need to solve #65.

peastman:
Depending on the details of your model, you might need to use `torch.jit.trace()` instead of `torch.jit.script()`, which also requires providing example inputs.

raimis:
Yes, both options should be available:
```python
force = TorchForce('module.pt')
force = TorchForce(module)
```

In practice, I have never needed `torch.jit.trace()`.

peastman:
This ties in with #65 and #95.  It was suggested to make serialization include the full model in the XML output.  The problem is that a TorchForce object doesn't actually store a model.  All it stores is a filename, which gets loaded each time you create a Context.  So that would involve storing information in the XML that isn't actually present in the TorchForce object, leading to a variety of problems.

It could be useful to allow the TorchForce to actually store the model.  We could add a constructor that takes a `torch::jit::script::Module` in the C++ API, or a `torch.jit.ScriptModule` in the Python API.  There would likewise be a `getModule()` method to retrieve it.  It would get stored in XML during serialization.

In the code above, the force creation would become

```python
force = TorchForce(torch.jit.script(module))
```

or if you need to use tracing,

```python
force = TorchForce(torch.jit.trace(module, exampleInputs))
```

cc @raulppelaez

RaulPPelaez:
What should be done about the file name in this case?
It seems redundant to have it as a property of TorchForce if a ```getModule()``` member exists. An instance constructed with a module directly does not have a notion of a file name. 

On the other hand ```TorchForce(string)``` can be implemented by calling ```TorchForce(torch::jit::script::module)```. The point being in both cases ```getModule()``` makes sense, while ```getFile()``` kind of looses meaning.

I propose the following solutions to this, trying to maintain retrocompatibility by not removing ```getFile()```. Calling ```getFile()``` could either:
[...]

Issue #65, opened 2022-02-11
Serialization of `TorchForce` is not self-contained


raimis:
When serializing `TorchForce`, the file name of the PyTorch model is save rather its content:
```python
from openmmtorch import TorchForce
from openmm import XmlSerializer

force = TorchForce('filename.pt')
print(XmlSerializer.serialize(force))
```
```xml
<?xml version="1.0" ?>
<Force file="filename.pt" forceGroup="0" outputsForces="0" type="TorchForce" usesPeriodic="0" version="1">
	<GlobalParameters/>
</Force>
```

So, it is a user responsibility to ensure that the right model file is in the right place during desalinization.

Ideally, the serialized object should be self-contained. We should save the file content, not it's name. Or, at least, we should store a hash of the file content and verify it during desalinization.

jchodera:
The [TorchScript format](https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/docs/serialization.md) is actually a ZIP file, so it is a binary format. We can simply `base64` encode/decode this file. It's [simple to implement in C++11](https://gist.github.com/williamdes/308b95ac9ef1ee89ae0143529c361d37), and there is a [proposed `std::string base64_encode/decode` addition](https://renenyffenegger.ch/notes/development/Base64/Encoding-and-decoding-base-64-with-cpp/index) that would further simplify this in future.

raimis:
@RaulPPelaez I guess, this was solved by #97 too.

PR #95, opened 2023-01-20
Make TorchForce serialization self-contiained

RaulPPelaez:
This pull request tries to solve #65 and eventually #66 
Currently TorchForce serialization includes only the filename of the model, not its contents.

- [x] Make the serializer record version 2
- - [x] Store the contents of the model file when serializing
- [x] Version 1 deserialization behaves as before.
- Version 2: 
 - [x] Serializes the filename provided by TorchForce for record 
  - [x] Serializes the file contents stored in base64
   - [x] Deserialization ignores file name and uses encoded contents
- - -[ ] Constructs the TorchForce by using a torch::jit::Module object. (Currently writes a temporal file, which is passed to TorchForce)
-----------
- [ ] Write a constructor for `TorchForce` that accepts a torch::jit::Module directly
- [ ] Modify the tests accordingly 
- [x] Add python-side tests for serialization of new functionality


peastman:
Thanks for the contribution!  Here are some thoughts based on a first read through.

The handling of the file version isn't quite correct.  It's not a property of the class (as you imply by adding a static field for it) nor of an object (as you imply when you refer to it as `this->version`).  Rather, it's a property of a particular file.  `serialize()` should hardcode the version number that it's writing out, i.e. `setIntProperty("version", 1)` instead of `setIntProperty("version", this->version)` and `deserialize()` should hardcode the version number it's looking for at each point in the code.

You have a block of code to implement the case version > 2, even though we've never defined any version other than 1?

The formatting is inconsistent with the rest of the code.  `if (...) {` needs spaces before `(` and after `)`.  Include spaces around the lowest precedence operator on each line, e.g. `valb -= 8;`.  Add a space after `#include`.  Indentation is by four spaces, not two.  The file specifies `using namespace std`, so you don't need the `std::` prefix.  The `not` operator is an archaic remnant from the days when not all keyboards had full sets of punctuation marks.  Use `!` instead.  Names should use capitalization, not underscores, to separate words.

Creating a separate `detail` namespace just for a few static functions is unnecessary.  They're already only visible from within the same file.  That's what static does.

Python tests should go in `python/tests`, not `serialization/tests`.

This won't work correctly:

```python
with tempfile.NamedTemporaryFile(delete=False) as temp:
    model_filename = temp.name
```

The temporary file gets deleted as soon as it exits the `with` block.  Then you keep around the name of that file, and later write a new file with that name which will not get deleted.

Code posted to Stack Overflow without any license specified is automatically covered by CC-BY-SA.  That's a viral license we shouldn't be using here.  We need a different implementation of base64 encoding that has an acceptable license, preferably either public domain, MIT, or BSD.

RaulPPelaez:
Thanks for your feedback, Peter! I will do my best to follow your formatting advice in the future. Does the codebase follow some standard formatting style (e.g llvm or google)? So I can configure a linter.
> The handling of the file version isn't quite correct.

By looking at other SerializationProxy classes in OpenMM, I now see that serialize() writes some hardcoded version and deserialize() tries to parse the widest range of versions possible. Please correct me if I am wrong.

> You have a block of code to implement the case version > 2, even though we've never defined any version other than 1?

I designed so a file containing version 1 only expects a file name for the model (the current behavior) and, if an encoded model file is also included, to check that the contents of the model file match the one encoded. Whereas a file containing a version number of 2 makes the Serializer expect an encoded model file and not a filename.


I clearly have some misconceptions about the version number. AFAIK, the final goal is to somehow modify the serializer, and TorchForce, to be able to work with an encoded model file instead of just a filename. 
This requires to modify the serialized file to hold different parameters (e.g. fileContents instead of fileName). It makes sense for me that this translates into a different version number for the serializer, so that fileName is something that should not be present if the version of the file is "1", etc. 
Your message makes me think that a new version is not the way to go about this. Could you give me some pointers? Should version 1 be able to handle both a fileName or the contents of one?



> The temporary file gets deleted as soon as it exits the `with` block. Then you keep around the name of that file, and later write a new file with that name which will not get deleted.

I understood delete=False made the file [not be deleted](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) after closure.
I have a pytest fixture at the end that deletes the files:

``` python
@pytest.fixture(autouse=True)
def cleanup():
    os.remove(serialized_filename)
    os.remove(model_filename)
```
The tests work as I would expect, but maybe I am abusing some undefined behavior without knowing, I will investigate further.
What would be a better approach to get a name for a temporary file?

peastman:
The purpose of the version number is to maintain backward compatibility.  Even when we change the format to support new features, new versions should still be able to read files created by older versions.  It also defines the limits of forward compatibility.  If a file was created by a new version, and you try to load it with an older version that can't support some of its features, it needs to realize that.  It should throw an exception rather than silently loading the file incorrectly.

The filename is part of the public API for TorchForce.  You can query it with `getFile()`.  That means the proxy will always need to record it.  If nothing else, it serves as a record of what model you created that force from.

This PR changes the behavior in some cases.  Currently, if you create a TorchForce from a model file, serialize it, and then reload it, it will use whatever model is in the file at load time, even if that's different from what was in that file when you originally created the force.  That probably isn't the best behavior, and it's preferable for it to use the same model it originally had.  So changing it is ok, but we need to make sure the behavior is consistent and predictable.  You shouldn't get one behavior if a particular file happens to be in your current working directory, but a different behavior if it isn't.  I suggest this.

- It should write the version number as 2.
- It should write both the filename and the model.
- When reading a file,
   - if the version is 1, load the model from the file specified in the filename.
   - if the version is 2, use the serialized model.

When loading this file with an older version, it will see that the version number is newer than it supports and throw an exception.  That is the correct behavior: it can't interpret the file in the way that was intended when it was written.

Regarding the temporary files, the intended use of NamedTemporaryFile is that you put everything involving it inside the `with` block.  Inside the block you write the file and do anything you need with it.  Once you exit the block the file is automatically deleted and you never look at it (or the filename) again.

RaulPPelaez:
Thanks, Peter. I think I addressed all your comments in my latest commits.
OpenSSL is a current dependency and has a base64 encoder/decoder, so I switched to using that.

Before the next review, I have a couple of questions:
1- Doing the following:
``` python
class ForceModule(torch.nn.Module):
    """A simple module that can be serialized"""
    def forward(self, positions):
        return torch.sum(positions**2)


```

peastman:
OpenSSL is *not* a dependency.  We shouldn't add it as one.

peastman:
I think we need to pause and consider the design.  Serialization is supposed to exactly encode the information content of a single object, nothing more and nothing less.  Currently, a TorchForce does not store the model, only a reference to the file name.  Therefore, the serialized file should contain only the filename and nothing else.  We're trying to make it store more than that, but without actually adding more information to the TorchForce object.  That's leading to logical inconsistencies.

If the serialized file is going to store the complete model, then TorchForce needs to store the complete model as well.  That requires a constructor that accepts a model rather than just a filename, and a public method for retrieving the model.  That adds a lot of complexity.

Do we really want that?  Can we enumerate the use cases we're trying to address?  What are the costs and benefits of that design compared to the current design?

RaulPPelaez:
I see. I will wait on this, then.

RaulPPelaez:
Work on this has been moved to #97, I am closing this pull request.

PR #96, opened 2023-01-24
Attempt at fixing SWIG issues

peastman:
Let's see if this works.


peastman:
I think that fixed it?  Really we should put the include into OpenMMSwigHeaders.i so each downstream package doesn't need to do it.  This will do for the moment though.

raimis:
@peastman is something more to do here? The tests are passing, so it should be good.

peastman:
I'll merge it once the OpenMM 8.0 release is out.

PR #97, opened 2023-01-30
Add a constructor to TorchForce that takes a torch::jit::Module

RaulPPelaez:
With the changes in this pull request ```TorchForce(const string& fileName)``` is implemented by delegating to a new constructor ```TorchForce(const torch::jit::Module&)```.
On construction ```TorchForce``` now calls ```torch::jit::load``` which introduced, C++ side, the restriction that when using the string constructor the file has to exists. 

Previously ```TorchForce``` could be created by providing a non-existent module file name. This was being abused in the C++ serialization test, which I had to modify accordingly. 
 I also added a comparison check for the module itself (in addition to the file name) to this test.

When the instance is constructed with a module directly, ```TorchForce::getFile()``` will return an empty string.
Regardless of the chosen constructor, ```TorchForce``` holds both a file name (even if empty) and a model.

Related to #66 .


RaulPPelaez:
It is not possible yet to use this constructor from the Python side. I do not know how to expose this constructor to Python, which entails telling swig how to transform the object returned by ```torch.jit.script(model)``` in Python to the equivalent ```torch::jit::Module``` in C++.
As far as I understand from the docs, the only way to do this is to serialize the model in Python and deserialize it in C++.
But again, I do not know how to tell swig to do that.
Any suggestions about this?

peastman:
[This page](https://discuss.pytorch.org/t/how-to-get-the-computation-graph-and-pass-it-to-c/114754) gives an example of passing a PyTorch module from Python to C++.

A complication is that PyTorch uses pybind11 to interface between Python and C++, while OpenMM uses SWIG.  [This page](https://github.com/pybind/pybind11/issues/1706) has some links to examples of interfacing between pybind11 and SWIG.

openmmapi/src/TorchForce.cpp
```diff
@@ -34,18 +34,28 @@
...
 
 using namespace TorchPlugin;
 using namespace OpenMM;
 using namespace std;
 
-TorchForce::TorchForce(const std::string& file) : file(file), usePeriodic(false), outputsForces(false) {
+TorchForce::TorchForce(const torch::jit::Module& module) : file(), usePeriodic(false), outputsForces(false), module(module) {
+}
+
+TorchForce::TorchForce(const std::string& i_file) : TorchForce(torch::jit::load(i_file)) {
```
peastman, line 47: The argument name should be `file`, not `i_file`.

peastman:
I posted on the [PyTorch discussion forum](https://discuss.pytorch.org/t/pass-model-to-c-wrapped-with-swig/171405) asking for help about how to create the wrapper.

RaulPPelaez:
Thanks Peter, I actually posted a [pybind11 version of your question](https://discuss.pytorch.org/t/passing-result-of-torch-jit-script-to-c-function/171380) a few hours before you to the pytorch forums. I figured it would help me understand how to take it to swig later on, since I am familiar with pybind.

RaulPPelaez:
I found some functions to go from a PyObject* to a Module, but I am clueless SWIG-side.
Maybe you can help. I defined the following typemap in openmmtorch.i:
```c++
%typemap(in) torch::jit::Module {
  py::object o = py::reinterpret_borrow<py::object>($input);
  torch::jit::Module module = torch::jit::as_module(o).value();
  $1 = module;
}
```
This compiles, but my typemap is clearly not being used, as I get a warning:
```
openmm-torch/python/openmmtorch.i:40: Warning 472: Overloaded method TorchPlugin::TorchForce::TorchForce(torch::jit::Module) with no explicit typecheck typemap 
for arg 0 of type 'torch::jit::Module'   
```
Then trying to run from Python:
```python
>>> import openmmtorch as ot; import torch as pt; ot.TorchForce(pt.jit.load("forces.pt"))
Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/shared/raul/mambaforge/envs/openmm/lib/python3.9/site-packages/openmmtorch.py", line 70, in __init__
    _openmmtorch.TorchForce_swiginit(self, _openmmtorch.new_TorchForce(*args))
TypeError: Wrong number or type of arguments for overloaded function 'new_TorchForce'.
  Possible C/C++ prototypes are:
    TorchPlugin::TorchForce::TorchForce(std::string const &)
    TorchPlugin::TorchForce::TorchForce(torch::jit::Module)
```
(I change the constructor to take an instance instead of a const ref to play around with this)

RaulPPelaez:
Adding a typecheck like this  does the trick:
```
%typecheck(SWIG_TYPECHECK_POINTER) torch::jit::Module {
  py::object o = py::reinterpret_borrow<py::object>($input);
  $1 = torch::jit::as_module(o).has_value()?1:0;
}
```

peastman:
I think you need to write it for ScriptModule, not Module?

RaulPPelaez:
See is the definition of the [as_module](https://github.com/pytorch/pytorch/blob/438f12d91a8adec94145b51181db310dccb4e652/torch/csrc/jit/python/module_python.h#L12-L19) function in torch.
I believe torch::jit::Module corresponds to ScriptModule

peastman:
`ScriptObject` doesn't appear anywhere in the API docs.  It looks like it's `torch._C.ScriptObject`, which subclasses `pybind11_builtins.pybind11_object`.  That suggests it's an internal class created by pybind11, which isn't supposed to be visible to users.

In the typemap

```c++
%typemap(out) const torch::jit::Module&{
    py::object o = py::cast(const_cast<torch::jit::Module*>($1));
    $result = o.release().ptr();
}
```

is there some way of telling pybind11 what Python class the result should be represented as?

RaulPPelaez:
I have been researching and my conclusion is that there is simply [no way](https://discuss.pytorch.org/t/passing-python-torch-nn-module-to-c-torch-module/89684/2) to go from C++ torch::jit::Module to the corresponding Python type. Although undocumented, [the opposite indirection is possible](https://github.com/pytorch/pytorch/blob/438f12d91a8adec94145b51181db310dccb4e652/torch/csrc/jit/python/module_python.h#L12-L19).
The functionality is just not there (and the documentation for the C++ python bindings of torch is really scarce).
If we adhere to what is documented and part of the API there is [only one way to do this](https://pytorch.org/docs/stable/notes/serialization.html#serializing-torch-nn-modules-and-loading-them-in-c), serialize the module in C++ and deserialize in Python. Pybind allows to plug this into a swig typecheck, like this:
```c++
%typemap(out) const torch::jit::Module&{
  auto fileName = std::tmpnam(nullptr);
  $1->save(fileName);
  $result = py::module::import("torch.jit").attr("load")(fileName).release().ptr();
  std::remove(fileName);
}
```
With this it is now possible to do the following
```python
import openmmtorch as ot
import torch
m=torch.jit.load("forces.pt")
f=ot.TorchForce(m)
m2= f.getModule()
f2 = ot.TorchForce(m2)
print(m)
print(m2)
print(m==m2)
print(m.save_to_buffer() == m2.save_to_buffer())
```
Which outputs:
```python
RecursiveScriptModule(original_name=Forces)
RecursiveScriptModule(original_name=Forces)
False
True
```
I believe this is a sensible approach. There is only one downside I can think of, which is the fact that you cannot modify the module being used by TorchForce after its creation, i.e ```getModule()``` gives you a clone, not a reference.
I am not sure, however, if the desired API seeks to guarantee this. 
What do you think?

peastman:
That sounds like a good solution.  I think we want TorchForce to be immutable in any case: once you create it, the module never changes.  In C++ `getModule()` returns a const reference, which is consistent with that behavior.  Having the Python version return a copy is also consistent.

peastman:
Can we add a few sentences to the description at the top of TorchForce describing the new behavior?  Something like this.

> The PyTorch module can either be passed directly as an argument to the constructor or loaded from a file.  In either case, the constructor makes a copy of the module in memory.  Later changes to the original module or to the file do not affect it.

peastman:
Tests are passing.  This is looking really good.  I think the only major thing we're missing is the one we started from: updating the serialization proxy to record the module.

RaulPPelaez:
In fact currently Context initialization fails python side when using the new constructor. Context tries to serialize and deserialize, but since currently serialization writes an empty string as fileName deserialization fails with an "Unrecognized format" error. I added some tests to showcase this, which should pass once serialization is taken care of.

RaulPPelaez:
I merged my commits from #95. Now all serialization tests are passing and the issue from my previous message is fixed.

I think the only dangling issue now is dealing with the base64 encoding. In my current commits the functions `EVP_EncodeBlock` and `EVP_DecodeBlock` from OpenSSL are used, but you wrote that adding OpenSSL as an explicit dependency was not desirable.
Can you think of some alternative? I do not think rolling a new encoding function from scratch is a good idea. It is a tricky algorithm already implemented by many other projects.

RaulPPelaez:
What about using hex encoding instead of base64? It is more lengthy, but not that it matters. Here is a possible implementation:

```c++
using namespace std;

string hexEncode(const string &input) {
  stringstream ss;
  ss << hex << setfill('0');
  for (const unsigned char &i : input) {
    ss << setw(2) << static_cast<uint64_t>(i);
  }
  return ss.str();
}

string hexDecode(const string &input) {
  string res;
  res.reserve(input.size() / 2);
  for (size_t i = 0; i < input.length(); i += 2) {
    istringstream iss(input.substr(i, 2));
    uint64_t temp;
    iss >> hex >> temp;
    res += static_cast<unsigned char>(temp);
  }
  return res;
}

string hexEncodeFromFileName(const string &filename) {
  ifstream input_file(filename, ios::binary);
  stringstream input_stream;
  input_stream << input_file.rdbuf();
  return hexEncode(input_stream.str());
}

int main() {
  string encoded = hexEncodeFromFileName("input.txt");
  auto decoded = hexDecode(encoded);
  ofstream("output.txt", ios::binary) << decoded;
  ifstream input_file("input.txt", ios::binary);
  stringstream input_stream;
  input_stream << input_file.rdbuf();
  assert(input_stream.str() == decoded);
  return 0;
}
```
Its just a few lines of standard C++ library calls. Seems to work with the example model files in the repo.

peastman:
Hex seems like a fine solution.  Models are unlikely to be so large that the extra space is an issue.

RaulPPelaez:
Then I would say all functionality has been included and tested, this is ready to merge. Please review.

python/tests/TestSerializeTorchForce.py
```diff
@@ -0,0 +1,50 @@
...
+        fileContents = f.read()
+    return fileContents
+
+def deserialize(filename):
+    other_force = XmlSerializer.deserialize(readXML(filename))
+
+def test_serialize(temporal_path):
+    model_filename = temporal_path + "/model.pt"
+    serialized_filename = temporal_path+ "/stored.xml"
+    createAndSerialize(model_filename, serialized_filename)
```
peastman, line 44: I think you mean "temporary", not "temporal"?  (Very different meanings!)  Anyway, this is a cleaner way of doing it:

```python
import os
import tempfile
...
with tempfile.TemporaryDirectory() as tempdir:
    model_filename = os.path.join(tempdir, 'model.pt')
```

Advantages are that 1) it handles Windows path separators correctly, 2) the directory is deleted as soon as you're done with it, 3) it's less code, and 4) it doesn't use pytest-specific features, in case we switch to a different test framework in the future.

serialization/src/TorchForceProxy.cpp
```diff
@@ -32,19 +32,60 @@
...
     node.setStringProperty("file", force.getFile());
+    try {
+        auto tempFileName = std::tmpnam(nullptr);
+        force.getModule().save(tempFileName);
+        node.setStringProperty("encodedFileContents", hexEncodeFromFileName(tempFileName));
+        std::remove(tempFileName);
+    }
+    catch (...) {
+        throw OpenMMException("Could not serialize model file.");
+    }
```
peastman, line 88: Can we provide more information to the user?  Something like this:

```c+
catch (exception& ex) {
    throw OpenMMException("TorchForceProxy: Could not serialize model file: "+ex.what());
}

File: serialization/src/TorchForceProxy.cpp [hunk 4]
```diff
@@ -55,9 +96,24 @@ void TorchForceProxy::serialize(const void* object, SerializationNode& node) con
...
-    TorchForce* force = new TorchForce(node.getStringProperty("file"));
+    TorchForce* force;
+    if (storedVersion == 1) {
+        string fileName = node.getStringProperty("file");
+        force = new TorchForce(fileName);
+    } else if (storedVersion == 2) {
+        const string storedEncodedFile = node.getStringProperty("encodedFileContents", "");
+        if (storedEncodedFile.empty()) {
+            throw OpenMMException("Found and empty model file.");
+        }
```
peastman, line 110: That message is a bit confusing, since there's no model file involved here.  I would remove the default value from `getStringProperty()`, which makes the `empty()` check unnecessary.  If this is version 2 or greater, the `encodedFileContents` property should be present.  If it isn't, the XML is invalid and you'll properly get an exception saying it's missing.

serialization/src/TorchForceProxy.cpp
```diff
@@ -55,9 +96,24 @@ void TorchForceProxy::serialize(const void* object, SerializationNode& node) con
...
-    if (node.getIntProperty("version") != 1)
+    int storedVersion = node.getIntProperty("version");
+    if (storedVersion > 2)
         throw OpenMMException("Unsupported version number");
-    TorchForce* force = new TorchForce(node.getStringProperty("file"));
+    TorchForce* force;
+    if (storedVersion == 1) {
+        string fileName = node.getStringProperty("file");
+        force = new TorchForce(fileName);
+    } else if (storedVersion == 2) {
```
peastman, line 106: I would just make this `else`, not `else if`.  This is presumably how all future versions will load it as well.

serialization/tests/TestSerializeTorchForce.cpp
```diff
@@ -34,18 +34,30 @@
...
-    // Create a Force.
+struct ExampleModuleImpl : torch::nn::Module { };
+TORCH_MODULE(ExampleModule);
+
+string writeExampleModuleFile() {
+    auto fileName = string(tmpnam(nullptr)) + ".pt";
+    ExampleModule module;
+    torch::save(module, fileName);
+    return fileName;
+}
```
peastman, line 55: What are these changes for?  Why go to the work to build a new module and write it to disk, rather than just using `module.pt` which is already present?

peastman:
This is looking really good.  I made a few suggestions, but they're all very minor ones.

serialization/src/TorchForceProxy.cpp
RaulPPelaez: Great! I was not aware of getStringProperty throwing.

serialization/tests/TestSerializeTorchForce.cpp
RaulPPelaez: I modified these tests to try both the fileName and the jit Module constructors using an already available module

RaulPPelaez:
Thanks Peter, I agree with your suggestions and addressed them in the latest commits.

peastman:
I think this is ready to merge.  Thanks for the really nice new feature!

PR #101, opened 2023-03-24
Make TorchForce compatible with CUDA graphs

RaulPPelaez:
This is a continuation of the work started in #68 
This PR introduces two new functionalities:
1. The constructors of TorchForce now allow a map containing a series of options ( currently only one for enabling cuda graphs)
2. The CUDA Platform can now build the model into a CUDA graph.
 
Tasks:
-[ ] Document the CUDA graph functionality and the new constructors
-[ ] Serialization new constructors
-[ ] Add test for new constructors
-[ ] Include options in serialization


peastman:
Interesting.  Can you give some background on the design, and describe what you changed and why?  There are a lot of differences, some of which I can't tell what they're for.  That includes minor edits to comments in seemingly unrelated code, removing the context management from `initialize()`, etc.  If you can revert changes that aren't directly related to this feature, it would make the PR easier to review.

We should give some thought to the mechanism for enabling graphs.  The property mechanism you created is one possibility.  We should consider others as well: allow the model to directly report whether CUDA graphs should be used, provide a specific function for enabling them, etc.  If we do go with the property mechanism, we need to document what the allowed properties are and provide a way to programmatically list the properties of an object.

RaulPPelaez:
I tried to pick up where raimis left the PR by merging with current master, but there are a lot of conflicts and unrelated changes as you said. Let me just close this and reopen so we can discuss anew.

PR #104, opened 2023-03-29
Document the constructor that takes a scripted module

RaulPPelaez:
I noticed the option to construct TorchForce using a jit.script module directly ( #97 ) was not documented in the README.md. This PR fixes it.


RaulPPelaez:
Well, at least we can leave it here as a reminder then :P

RaulPPelaez:
This could be merged now that the functionality is released

PR #106, opened 2023-04-20
Torch2 compatibility

RaulPPelaez:

PR #109, opened 2023-05-18
[DNM] Test torch2 on gpu

mikemhenry:


mikemhenry:
Tests PR #106

RaulPPelaez:
Tests are failing here because this PR appears to somewhat start from the wrong git commit. Files like TorchForce.h and the SWIG typemaps are versions older than the current master but some python tests seem to be the latest ones. The segfaults you get are because latest tests expect TorchForce constructor to take up to two arguments, but the TorchForce.h in your PR only has one.

mikemhenry:
Thanks! Superseded by https://github.com/openmm/openmm-torch/pull/110

PR #114, opened 2023-07-27
Fix test file paths

raimis:
Upstream a patch from https://github.com/conda-forge/openmm-torch-feedstock/pull/36 to fix the test file paths.


raimis:
We need to unify the directory structure for the CI test and the build tests.

peastman:
This was fixed in #143.

PR #116, opened 2023-09-04
Clone the module on TorchForceImpl initialization

RaulPPelaez:
This avoids different contexts sharing the same instance of the module owned by TorchForce.
Fixes https://github.com/openmm/NNPOps/pull/113 and https://github.com/openmm/NNPOps/issues/112


peastman:
Looks good.

raimis (approved):

PR #122, opened 2023-09-26
Use the same stream as the OpenMM context when replaying a CUDA graph.

RaulPPelaez:
This issue https://github.com/torchmd/torchmd-net/issues/220 unveiled a race condition in the current graph replay mechanism. By asking torch for a CUDA stream we risk a later OpenMM workload to run on a different one and thus access non-ready data.

In this PR I make sure that the CUDA graph replay uses the same stream as the OpenMM context, while ensuring that CUDA graph capture occurs in a non-default stream (a requirement for capture).

@raimis please confirm this fixes the original issue in your environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ torch_force = TorchForce('model.pt', {'useCUDAGraphs': 'true'})
 The first time the model is run, it will be compiled (also known as recording) into a CUDA graph. Subsequent runs will use the compiled graph, which can be significantly faster. It is possible that compilation fails, in which case an `OpenMMException` will be raised. If that happens, you can disable CUDA graphs and try again.
 
 It is required to run the model at least once before recording, in what is known as warmup.
-By default ```TorchForce``` will run the model just once before recording, but longer warming up might be desired. In these cases one can set the property ```CUDAGraphWarmupSteps```:
+By default ```TorchForce``` will run the model just a few times before recording, but controlling warmup steps might be desired. In these cases one can set the property ```CUDAGraphWarmupSteps```:
 ```python
 torch_force.setProperty("CUDAGraphWarmupSteps", "12")
 ```
--- a/openmmapi/src/TorchForce.cpp
+++ b/openmmapi/src/TorchForce.cpp
@@ -42,7 +42,7 @@ using namespace OpenMM;
 using namespace std;
 
 TorchForce::TorchForce(const torch::jit::Module& module, const map<string, string>& properties) : file(), usePeriodic(false), outputsForces(false), module(module) {
-    const std::map<std::string, std::string> defaultProperties = {{"useCUDAGraphs", "false"}, {"CUDAGraphWarmupSteps", "1"}};
+    const std::map<std::string, std::string> defaultProperties = {{"useCUDAGraphs", "false"}, {"CUDAGraphWarmupSteps", "10"}};
     this->properties = defaultProperties;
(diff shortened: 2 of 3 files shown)


RaulPPelaez:
Sadly this was not enough and the error popped out again. 
I had to also increase the warmup steps to at least 5:
```python
            force.setProperty("CUDAGraphWarmupSteps", "5")
```
I have no idea why 5 is the magic number. Requiring more than 1 (the default) does not surprise me (for instance, [DDP requires 11](https://pytorch.org/docs/master/notes/cuda.html#usage-with-distributeddataparallel))

The error that is generated makes me think warmup is indeed the issue here, this warning goes away when the reproducer runs.
```shell
[W manager.cpp:335] Warning: FALLBACK path has been taken inside: runCudaFusionGroup. This is an indication that codegen Failed for some reason.                             
To debug try disable codegen fallback path via setting the env variable `export PYTORCH_NVFUSER_DISABLE=fallback`                                                            
 (function runCudaFusionGroup)   
```
I guess torch is sometimes choosing a non-graphable path during the first calls. It sounds like torch is waiting for a few calls to try and pass the model through nvfuser. By CUDA graphing early we break this.

Not sure how to go about this, perhaps increasing the default warmup steps to something like 10 would be sensible?
I feel like we really do not have control over things like torch calling nvfuser...

I could also try to rewrite it so that instead of TorchForce calling the torch model X times during initialization, it tracks the number of times it has been called and only captures after X calls. I am more inclined to leave it as is, since whatever torch does to warmup might only happen if there is no other code in between calls.

raimis:
@RaulPPelaez good work! I'll test the fix.

raimis:
@RaulPPelaez I can preliminarily confirm that the fix works. I'll run a complete simulation to verify definitely.

raimis:
> I could also try to rewrite it so that instead of TorchForce calling the torch model X times during initialization, it tracks the number of times it has been called and only captures after X calls. I am more inclined to leave it as is, since whatever torch does to warmup might only happen if there is no other code in between calls.

Yes, I agree. You can leave it as it is.

README.md
```diff
@@ -270,7 +270,7 @@ torch_force = TorchForce('model.pt', {'useCUDAGraphs': 'true'})
 The first time the model is run, it will be compiled (also known as recording) into a CUDA graph. Subsequent runs will use the compiled graph, which can be significantly faster. It is possible that compilation fails, in which case an `OpenMMException` will be raised. If that happens, you can disable CUDA graphs and try again.
 
 It is required to run the model at least once before recording, in what is known as warmup.
-By default ```TorchForce``` will run the model just once before recording, but longer warming up might be desired. In these cases one can set the property ```CUDAGraphWarmupSteps```:
+By default ```TorchForce``` will run the model just a few times before recording, but controlling warmup steps might be desired. In these cases one can set the property ```CUDAGraphWarmupSteps```:
```
raimis, line 273: a few times --> a few times (10)

raimis (approved):
Looks good! Just be explicit in the docs.

README.md
RaulPPelaez: I did not want to put a number because its easy to forget to change it if we change our minds in the future, so I wanted to be vague on purpose -.- Almost forgot myself this time.

peastman (approved):
Looks good.

README.md
raimis: OK!

PR #142, opened 2024-05-15
Support parameter derivatives

### LINKED ISSUE #140: How to use global parameter and PBC at the same time?
Hi, 

I was trying to add a global parameter to my torchForce, but I could not figure out how to make it work along with the PBC. Could you help me have a look?

Here is the structure of my torchForce class:
```
class ForceModule(torch.nn.Module):
def __init__(self):
    ...
def forward(self, coordinates: torch.Tensor, cell: torch.Tensor, scale: float, RcNL: float = 10.0):
     return energy
```

Then when I try to load the model with 
```
force = TorchForce('model.pt')
force.setUsesPeriodicBoundaryConditions(True)
force.addGlobalParameter('scale', 0.9)
```
and try to run a MD with it,  I get the following error:
```
Traceback (most recent call last):
  File "/home/xie1/torchSANI_jit/scripts/openmm_test_3.py", line 128, in <module>
    state = simulation.context.getState(getEnergy=True, groups={10})
  File "/home/xie1/psi4conda/envs/nnpops_env/lib/python3.10/site-packages/openmm/openmm.py", line 12111, in getState
    state = _openmm.Context_getState(self, types, enforcePeriodicBox, groups_mask)
openmm.OpenMMException: forward() Expected a value of type 'float' for argument 'scale' but instead found type 'Tensor'.
Position: 3
Declaration: forward(__torch__.___torch_mangle_0.ForceModule self, Tensor coordinates, Tensor cell, float scale, float RcNL=10.) -> Tensor
```

Thank you so much!

### LINKED ISSUE #141: Support parameter derivatives
TorchForce should have a `addEnergyParameterDerivative()` function so you can compute derivatives with respect to global parameters.  See #140.

RaulPPelaez:
Allows to get energy derivatives with respect to global parameters in TorchForce as in the following example:
```python
class ForceWithParameters(pt.nn.Module):

    def __init__(self):
        super(ForceWithParameters, self).__init__()

    def forward(
        self, positions: Tensor, parameter1: Tensor, parameter2: Tensor
    ) -> Tensor:
        x2 = positions.pow(2).sum(dim=1)
        u_harmonic = ((parameter1 + parameter2**2) * x2).sum()
        return u_harmonic


def example():
    # Create a random cloud of particles.
    numParticles = 10
    system = mm.System()
    positions = np.random.rand(numParticles, 3)
    for _ in range(numParticles):
        system.addParticle(1.0)

    # Create a force
    pt_force = ForceWithParameters()
    model = pt.jit.script(pt_force)
    tforce = ot.TorchForce(model)
    # Add some parameters
    parameter1 = 1.0
    parameter2 = 1.0
    force.setOutputsForces(False)
    force.addGlobalParameter("parameter1", parameter1)
    force.addEnergyParameterDerivative("parameter1")
    force.addGlobalParameter("parameter2", parameter2)
    force.addEnergyParameterDerivative("parameter2")
    # Enable energy derivatives for the parameter
    system.addForce(force)
    # Compute the forces and energy.
    integ = mm.VerletIntegrator(1.0)
    platform = mm.Platform.getPlatformByName(platform)
    context = mm.Context(system, integ, platform)
    context.setPositions(positions)
    state = context.getState(
        getEnergy=True, getForces=True, getParameterDerivatives=True
    )
    # See if the energy and forces and the parameter derivative are correct.
    # The network defines a potential of the form E(r) = (parameter1 + parameter2**2)*|r|^2
    r2 = np.sum(positions * positions)
    expectedEnergy = (parameter1 + parameter2**2) * r2
    assert np.allclose(
        r2,
        state.getEnergyParameterDerivatives()["parameter1"],
    )
    assert np.allclose(
        2 * parameter2 * r2,
        state.getEnergyParameterDerivatives()["parameter2"],
    )

```

Closes #140
Closes #141

### DIFF
--- a/platforms/cuda/src/CudaTorchKernels.cpp
+++ b/platforms/cuda/src/CudaTorchKernels.cpp
@@ -66,10 +66,10 @@ void CudaCalcTorchForceKernel::initialize(const System& system, const TorchForce
     outputsForces = force.getOutputsForces();
     for (int i = 0; i < force.getNumGlobalParameters(); i++)
         globalNames.push_back(force.getGlobalParameterName(i));
-    for (int i = 0; i < force.getNumEnergyParameterDerivatives(); i++){
+    for (int i = 0; i < force.getNumEnergyParameterDerivatives(); i++) {
         auto name = force.getEnergyParameterDerivativeName(i);
         energyParameterDerivatives.push_back(name);
-	cu.addEnergyParameterDerivative(name);
+        cu.addEnergyParameterDerivative(name);
     }
     int numParticles = system.getNumParticles();
 
@@ -86,6 +86,7 @@ void CudaCalcTorchForceKernel::initialize(const System& system, const TorchForce
     boxTensor = torch::empty({3, 3}, options);
     energyTensor = torch::empty({0}, options);
     forceTensor = torch::empty({0}, options);
+    gradientTensors.resize(force.getNumEnergyParameterDerivatives(), torch::empty({0}, options));
     // Pop the PyToch context
     CUcontext ctx;
     CHECK_RESULT(cuCtxPopCurrent(&ctx), "Failed to pop the CUDA context");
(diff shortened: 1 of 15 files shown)


RaulPPelaez:
@peastman, I do not know how to connect CustomCVForce with the parameter derivative functionality in TorchForce.
I am writing this in the new test
```python
    force = ot.TorchForce(model, {"useCUDAGraphs": "false"})
    # Add a parameter
    parameter = 1.0
    force.addGlobalParameter("parameter", parameter)
    force.setOutputsForces(False)
    if use_cv_force:
        # Wrap TorchForce into CustomCVForce
        force = mm.CustomCVForce("force")
        force.addCollectiveVariable("force", force)
    # Enable energy derivatives for the parameter
    force.addEnergyParameterDerivative("parameter")
    system.addForce(force)
```
For `use_cv_force=True` I get
```shell
>       force.addEnergyParameterDerivative("parameter")

../python/tests/TestParameterDerivatives.py:48: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <openmm.openmm.CustomCVForce; proxy of <Swig Object of type 'OpenMM::CustomCVForce *' at 0x7f010153b6f0> >, name = 'parameter'

    def addEnergyParameterDerivative(self, name):
        r"""
        addEnergyParameterDerivative(self, name)
        Request that this Force compute the derivative of its energy with respect to a global parameter. The parameter must have already been added with addGlobalParameter().
    
        Parameters
        ----------
        name : string
            the name of the parameter
        """
>       return _openmm.CustomCVForce_addEnergyParameterDerivative(self, name)
E       openmm.OpenMMException: addEnergyParameterDerivative: Unknown global parameter 'parameter'
```

RaulPPelaez:
There is a corner case I do not know how to handle.
It is possible that the model computes and provides the forces by calling autograd on the energy.
If the model does this by calling grad directly (instead of using energy.backwards) and sets retain_graph=False, then torch will not let you call backwards again.
In other words, this model is inherently incompatible with parameter derivatives:
```python
class EnergyForceWithParameters(pt.nn.Module):

    def __init__(self):
        super(EnergyForceWithParameters, self).__init__()

    def forward(
        self, positions: Tensor, parameter1: Tensor, parameter2: Tensor
    ) -> Tuple[Tensor, Tensor]:
        positions.requires_grad_(True)
        x2 = positions.pow(2).sum(dim=1)
        u_harmonic = ((parameter1 + parameter2**2) * x2).sum()
        # This way of computing the forces forcefully leaves out the parameter derivatives
        grad_outputs: List[Optional[Tensor]] = [pt.ones_like(u_harmonic)]
        dy = pt.autograd.grad(
            [u_harmonic],
            [positions],
            grad_outputs=grad_outputs,
            create_graph=False,
            retain_graph=False,  # This cannot be False if parameter derivatives are needed
        )[0]
        assert dy is not None
        forces = -dy
        return u_harmonic, forces
```
TorchMD-Net does exactly this. Not sure what to do about it.

platforms/cuda/src/CudaTorchKernels.cpp
```diff
@@ -238,17 +271,26 @@ double CudaCalcTorchForceKernel::execute(ContextImpl& context, bool includeForce
...
     // Get energy
     const double energy = energyTensor.item<double>(); // This implicitly synchronizes the PyTorch context
+    // Store parameter energy derivatives
+    auto& derivs = cu.getEnergyParamDerivWorkspace();
+    for (int i = 0; i < energyParameterDerivatives.size(); i++) {
+        // Compute the derivative of the energy with respect to this parameter.
+        // The derivative is stored in the gradient of the parameter tensor.
+        double derivative = gradientTensors[i].item<double>();
+        auto name = energyParameterDerivatives[i];
+        derivs[name] = derivative;
```
RaulPPelaez, line 292: Should I be summing here instead of overwritting?

peastman:
It looks like we both started working on this.  I'm almost finished with implementing it.

RaulPPelaez:
welp... Take what you want from here if you find anything useful.

peastman:
Really sorry for the confusion!  My implementation is at #141.  If you can review it, that would be great.

peastman:
Sorry, that should have been #143.

PR #143, opened 2024-05-15
Can compute parameter derivatives

peastman:
Implements #141.

### DIFF
--- a/openmmapi/src/TorchForce.cpp
+++ b/openmmapi/src/TorchForce.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2018-2022 Stanford University and the Authors.      *
+ * Portions copyright (c) 2018-2024 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors: Raimondas Galvelis, Raul P. Pelaez                           *
  *                                                                            *
@@ -110,6 +110,24 @@ double TorchForce::getGlobalParameterDefaultValue(int index) const {
 void TorchForce::setGlobalParameterDefaultValue(int index, double defaultValue) {
     ASSERT_VALID_INDEX(index, globalParameters);
     globalParameters[index].defaultValue = defaultValue;
+}
+
+int TorchForce::getNumEnergyParameterDerivatives() const {
+    return energyParameterDerivatives.size();
+}
+
+void TorchForce::addEnergyParameterDerivative(const string& name) {
+    for (int i = 0; i < globalParameters.size(); i++)
+        if (name == globalParameters[i].name) {
+            energyParameterDerivatives.push_back(i);
+            return;
+        }
+    throw OpenMMException(string("addEnergyParameterDerivative: Unknown global parameter '"+name+"'"));
+}
+
+const string& TorchForce::getEnergyParameterDerivativeName(int index) const {
+    ASSERT_VALID_INDEX(index, energyParameterDerivatives);
+    return globalParameters[energyParameterDerivatives[index]].name;
 }
 
 void TorchForce::setProperty(const std::string& name, const std::string& value) {
--- a/platforms/cuda/src/CudaTorchKernels.cpp
+++ b/platforms/cuda/src/CudaTorchKernels.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2018-2022 Stanford University and the Authors.      *
+ * Portions copyright (c) 2018-2024 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors: Raimondas Galvelis, Raul P. Pelaez                           *
  *                                                                            *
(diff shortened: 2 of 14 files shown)


RaulPPelaez:
We really need to fix the CI...

openmmapi/src/TorchForce.cpp
```diff
@@ -112,6 +112,24 @@ void TorchForce::setGlobalParameterDefaultValue(int index, double defaultValue)
...
 }
 
+int TorchForce::getNumEnergyParameterDerivatives() const {
+    return energyParameterDerivatives.size();
+}
+
+void TorchForce::addEnergyParameterDerivative(const string& name) {
+    for (int i = 0; i < globalParameters.size(); i++)
+        if (name == globalParameters[i].name) {
+            energyParameterDerivatives.push_back(i);
```
RaulPPelaez, line 122: I have seen this in other parts of OpenMM, what happens if I call this function twice with the same name? Is that handled somewhere before or after this?

platforms/cuda/src/CudaTorchKernels.cpp
```diff
@@ -173,26 +181,33 @@ void CudaCalcTorchForceKernel::addForces(torch::Tensor& forceTensor) {
...
+ * This function launches the workload in a way compatible with CUDA
+ * graphs as far as OpenMM-Torch goes.  Capturing this function when
+ * the model is not itself graph compatible (due to, for instance,
  * implicit synchronizations) will result in a CUDA error.
  */
 static void executeGraph(bool outputsForces, bool includeForces, torch::jit::script::Module& module, vector<torch::jit::IValue>& inputs, torch::Tensor& posTensor, torch::Tensor& energyTensor,
-                         torch::Tensor& forceTensor) {
+                         torch::Tensor& forceTensor, map<string, torch::Tensor>& derivInputs) {
+    vector<torch::Tensor> gradInputs;
+    if (!outputsForces)
```
RaulPPelaez, line 192: This should be `!outputsForces && includeForces` right?

platforms/cuda/src/CudaTorchKernels.cpp
```diff
@@ -173,26 +181,33 @@ void CudaCalcTorchForceKernel::addForces(torch::Tensor& forceTensor) {
...
         energyTensor = module.forward(inputs).toTensor();
         // Compute force by backpropagating the PyTorch model
         if (includeForces) {
             // CUDA graph capture sometimes fails if backwards is not explicitly requested w.r.t positions
-	    // See https://github.com/openmm/openmm-torch/pull/120/
-	    auto none = torch::Tensor();
-	    energyTensor.backward(none, false, false, posTensor);
-	    // This is minus the forces, we change the sign later on
+            // See https://github.com/openmm/openmm-torch/pull/120/
+            energyTensor.backward(none, false, false, gradInputs);
```
RaulPPelaez, line 209: This should also be called if the model does not return forces and they are also not required.

platforms/cuda/src/CudaTorchKernels.cpp
```diff
@@ -173,26 +181,33 @@ void CudaCalcTorchForceKernel::addForces(torch::Tensor& forceTensor) {
...
+            // This is minus the forces, we change the sign later on
             forceTensor = posTensor.grad().clone();
             // Zero the gradient to avoid accumulating it
             posTensor.grad().zero_();
```
RaulPPelaez, line 213: You need to zero the grads of the parameter derivatives I believe. Since the tensors are static they would get accumulated otherwise.

RaulPPelaez:
There should also be a Python side test. Take this one if you'd like:
https://github.com/openmm/openmm-torch/blob/4c9b5e509d15ce52db7600f7538fbe66538c1c2a/python/tests/TestParameterDerivatives.py

BTW, [this](https://github.com/openmm/openmm-torch/blob/4c9b5e509d15ce52db7600f7538fbe66538c1c2a/python/tests/TestParameterDerivatives.py#L44-L45) is still a problem. I do not think OpenMM-Torch can do anything about it, but perhaps there is a way to detect it and provide a useful message?

peastman:
Thanks for the comments.  It turned out that what I had written didn't work with CUDA graphs.  I restructured it to handle the input tensors in a different way and made the test case run both with and without graphs.  Can you see if it looks better now?

platforms/cuda/src/CudaTorchKernels.cpp
```diff
@@ -145,12 +151,17 @@ std::vector<torch::jit::IValue> CudaCalcTorchForceKernel::prepareTorchInputs(Con
...
-        inputs.push_back(torch::tensor(context.getParameter(name)));
-    return inputs;
+    for (const string& name : globalNames) {
+        // PyTorch requires us to set requires_grad to false before initializing a tensor.
+        globalTensors[name].set_requires_grad(false);
+        globalTensors[name][0] = context.getParameter(name);
+        if (paramDerivs.find(name) != paramDerivs.end())
+            globalTensors[name].set_requires_grad(true);
+        inputs.push_back(globalTensors[name]);
+    }
```
RaulPPelaez, line 164: I reckon you are resetting requires_grad as a replacement to zeroing the grads. I am surprised this works with CUDA graphs, since by doing this pytorch might be somehow freeing the gradient tensors and picking up new ones when reenabling.
I am weary you are depending on some UB by doing so.

peastman: I zero the grads at the end of `execute()`, just before returning.  This is because assigning a value to a tensor throws an exception if you don't set `require_grads` to false first.

peastman:
Is this ok to merge?  Getting CI working is its own major project.  I'm working on that in another PR.

PR #145, opened 2024-05-24
Add energy derivatives to README

RaulPPelaez:
We forgot to comment on energy derivatives in the README.
cc @peastman

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,17 +265,17 @@ The parameter derivatives can be queried by calling `getEnergyParameterDerivativ
 
 ```python
 import torch as pt
-from torch import Tensor
 from openmmtorch import TorchForce
 import openmm as mm
+
 
 class ForceWithParameters(pt.nn.Module):
 
     def __init__(self):
         super(ForceWithParameters, self).__init__()
 
-	def forward(self, positions: Tensor, k: Tensor) -> Tensor:
-		return k*torch.sum(positions**2)
+    def forward(self, positions: pt.Tensor, k: pt.Tensor) -> pt.Tensor:
+        return k * pt.sum(positions**2)
 
 
 numParticles = 10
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -256,6 +256,72 @@ to return forces.
 torch_force.setOutputsForces(True)
 ```
 
+Computing energy derivatives with respect to global parameters
+--------------------------------------------------------------
+
+Its possible to query `TorchForce` for the derivative of the energy with respect to global parameters. In order to do so the global parameters must be registered as energy derivatives. This is done by calling `addEnergyParameterDerivative()` for each parameter.
```
peastman, line 262: For the first sentence, how about, "TorchForce can compute derivatives of the energy with respect to global parameters."  Since you don't actually query the TorchForce to get them.

README.md
```diff
@@ -256,6 +256,72 @@ to return forces.
...
+
+    def __init__(self):
+        super(ForceWithParameters, self).__init__()
+
+    def forward(
+        self, positions: Tensor, parameter1: Tensor, parameter2: Tensor
+    ) -> Tensor:
+        x2 = positions.pow(2).sum(dim=1)
+        u_harmonic = ((parameter1 + parameter2**2) * x2).sum()
+        return u_harmonic
```
peastman, line 282: This seems overly complicated for an example;  One parameter would illustrate the point as well as two.  And I don't quite follow your code.  You sum over one dimension, multiply by a constant, and then sum over the remaining dimensions?

How about this:

```python
def forward(self, positions: Tensor, k: Tensor) -> Tensor:
    return k*torch.sum(positions**2)
```

(which is exactly the same as the earlier example for global parameters).

README.md
```diff
@@ -256,6 +256,72 @@ to return forces.
...
+
+
+def example():
```
peastman, line 285: All the other examples are much more minimal than this.  They don't give complete test cases with assertions, just the essential code for TorchForce and explanations of what it does.  In this case we need to explain how to create the force and register the parameter derivative:

```python
module = torch.jit.script(ForceModule())
force = TorchForce(module)
force.addGlobalParameter("k", 2.0)
force.addEnergyParameterDerivative("k")
```

and then how to query it from the Context:

```python
state = context.getState(getParameterDerivatives=True)
dEdk = state.getEnergyParameterDerivatives()["k"]
```

RaulPPelaez:
I addressed your comments, @peastman, please review again.

peastman:
That looks good.  Thanks!

PR #148, opened 2024-06-28
Freeze pytorch model

peastman:
Freezing the model allows PyTorch to do more optimizations on it.  I ran some tests on the Nutmeg models, and found typical speedups of about 2 to 7%.  Not huge, but worth doing.

There's another function called `optimize_for_inference()` that can do even more optimizations.  But the documentation says it's still in prototype and can sometimes slow models down.  For the moment, it seems best to stick to just freezing.


RaulPPelaez:
Thanks Peter, LGTM. Surprising that "eval" does not freeze already.

PR #152, opened 2024-08-19
Serialize property values

### LINKED ISSUE #151: Question about how to use CUDA graph
Hi, 

I would like to use the CUDA graph option to accelerate my simulation, and I was wondering is there anything I need to do when converting the model to torchscript and saving to a file in order to use CUDA graph? 

I tried simply adding `force.setProperty("useCUDAGraphs", "true")` to the simulation script but did not see any performance improvement compared to without CUDA graph. Is there any way I can investigate whether it is indeed using CUDA graph?

Thank you,
Xiaowei

peastman:
Fixes #151.  The property values weren't being serialized.


peastman:
I don't know why the Mac runner isn't starting.  I'll debug it in a separate PR.

PR #178, opened 2026-06-15
CI build for HIP

peastman:


.github/workflows/CI.yml
```diff
@@ -70,6 +85,13 @@ jobs:
         if: startsWith(matrix.os, 'ubuntu')
         uses: jlumbroso/free-disk-space@main
 
+      - name: "Install HIP 7 on Linux (if needed)"
+        if: matrix.hip-version == '7'
+        run: |
+          wget https://repo.radeon.com/amdgpu-install/7.2.4/ubuntu/noble/amdgpu-install_7.2.4.70204-1_all.deb
+          sudo apt install ./amdgpu-install_7.2.4.70204-1_all.deb
+          sudo apt update
```
timpalpant, line 83: I think an additional line is needed here to install ROCm. The `amdgpu-install_7.2.4.70204-1_all.deb` package they provide just adds some repositories to APT but doesn't install the libraries themselves: https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/quick-start.html#install-rocm
```suggestion
          sudo apt update
          sudo apt install rocm
```

peastman:
Some progress.  I have it building everything with HIP.  It then gets to this line which just reports what plugins could or couldn't be loaded:

> python -c "import openmm as mm; print('---Loaded---', *mm.pluginLoadedLibNames, '---Failed---', *mm.Platform.getPluginLoadFailures(), sep='\n')"

It prints the output (reference, CPU, and OpenCL all loaded, CUDA and HIP not loaded because of missing dependencies).  And then it crashes with a memory corruption error: `corrupted size vs. prev_size in fastbins`.  It hasn't actually run any HIP code.  All it's done is try and fail to load the HIP plugins because it couldn't find `libhiprtc.so.6`.  But something has still managed to corrupt memory.

PR #179, opened 2026-06-18
Create PythonTorchForce

peastman:
This adds PythonTorchForce.  It's very similar to PythonForce, except that the positions are passed as a Tensor instead of a State, and the forces are returned as a Tensor instead of a NumPy array.  When using the CUDA platform, this lets us skip some data copies between host and device, reducing overhead.

In principle we could do the same with HIP, but I haven't managed to compile anything against the HIP build of PyTorch.  See https://github.com/pytorch/pytorch/issues/181939.

So how much does it help?  I tried using AceFF 1.1 (a pretty fast potential as ML models go) to simulate a 2 nm water box with 774 atoms on a RTX 4080.  With the current PythonForce implementation, 100 time steps take 1.92 seconds.  Switching to PythonTorchForce reduces it to 1.88 seconds.  Oh well.  Looking at the profiler I can see that each invocation of the model takes about 18 ms.  In between there's a gap where it does data transfers and integration.  With PythonForce, that gap is about 1 ms.  PythonTorchForce reduces it to 0.5 ms.

So this is only going to be signficant for really fast models.  Still, it's something.


peastman:
There, I finally got all the tests to pass!  The CI workflow was very out of date.

.github/workflows/CI.yml
```diff
@@ -38,8 +38,8 @@ jobs:
             python-version: "3.11"
             pytorch-version: "2.4.*"
 
-          - name: MacOS Intel (Python 3.11, PyTorch 2.4)
-            os: macos-13
+          - name: MacOS Intel (Python 3.13, PyTorch 2.5)
```
epretti, line 43: The title of this job is updated but the specifications look the same?

README.md
```diff
@@ -5,364 +5,154 @@
...
+    return energy, forces
 ```
 
-Defining a model that uses periodic boundary conditions
--------------------------------------------------------
-
-When defining the model to perform a calculation, you may want to apply periodic boundary conditions.  
+The State contains global parameters and periodic box vectors.  The Tensor contains particle positions.  The function
+should compute the potential energy and forces, returning them as its two return values.  The energy should be a
+scalar Tenor containing the value in kJ/mol.  The forces should be a Tensor of shape `(# particles, 3)` containing
```
epretti:
```suggestion
scalar Tensor containing the value in kJ/mol.  The forces should be a Tensor of shape `(# particles, 3)` containing
```

python/openmmtorch.i
```diff
@@ -109,4 +120,226 @@ public:
...
+            for (size_t i = 0; i < input.length(); i += 2) {
+                std::istringstream iss(input.substr(i, 2));
+                uint64_t temp;
+                iss >> std::hex >> temp;
+                res.push_back(static_cast<unsigned char>(temp));
+            }
+            return res;
+        }
+
+        void serialize(const void* object, OpenMM::SerializationNode& node) const {
```
epretti, line 273: There was a recent issue about `PythonForce` not saving/restoring the force object's name.  `serialize()` and `deserialize()` here appear to need the same fixes.  The corresponding test should likewise check this.

python/tests/TestPythonTorchForce.py
```diff
@@ -0,0 +1,238 @@
...
+    def testMinimization(self):
+        """Test that PythonTorchForce works correctly with the minimizer."""
+        system = System()
+        for i in range(5):
+            system.addParticle(1.0)
+        force = PythonTorchForce(compute, {'k':2.5})
+        system.addForce(force)
+        positions = np.random.rand(5, 3)
+        integrator = VerletIntegrator(0.001)
+        context = Context(system, integrator, Platform.getPlatform('Reference'))
```
epretti: Can this test check the different platforms like the others?  The minimizer implementation is different on GPU now (and previously caused problems with PyTorch) so it would be good to make sure this is covered even if not on CI.

openmmapi/include/PythonTorchForce.h
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * This class provides a mechanism for computing forces and energy with Python code.  It is similar to the
+ * PythonForce class included with OpenMM, but it is specialized to give better performance when working with
+ * PyTorch models.
+ *
+ * To use it, define a Python function that takes two arguments: a State object and a Tensor of shape (# particles, 3).
+ * The State contains global parameters and periodic box vectors.  The Tensor contains particle positions.  The function
+ * should compute the potential energy and forces, returning them as its two return values.  The energy should be a
+ * scalar Tenor containing the value in kJ/mol.  The forces should be a Tensor of shape (# particles, 3) containing
```
epretti:
```suggestion
 * scalar Tensor containing the value in kJ/mol.  The forces should be a Tensor of shape (# particles, 3) containing
```

openmmapi/src/PythonTorchForceImpl.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+double PythonTorchForceImpl::calcForcesAndEnergy(ContextImpl& context, bool includeForces, bool includeEnergy, int groups) {
+    if ((groups&(1<<forceGroup)) != 0)
+        return kernel.getAs<CalcPythonTorchForceKernel>().execute(context, includeForces, includeEnergy);
+    return 0.0;
+}
+
+vector<string> PythonTorchForceImpl::getKernelNames() {
+    return {CalcCustomCPPForceKernel::Name()};
```
epretti, line 61: Should this be `CalcPythonTorchForceKernel::Name()`?

epretti:
The test failure looks strange...

```
Fatal Python error: Segmentation fault

Thread 0x00007f6cf76c7740 [pytest] (most recent call first):
  File "/home/runner/miniconda3/envs/build/lib/python3.14/site-packages/openmm/openmm.py", line 16692 in minimize
  File "/home/runner/work/openmm-torch/openmm-torch/python/tests/TestPythonTorchForce.py", line 190 in testMinimization
  ...

Current thread's C stack trace (most recent call first):
  ...
  Binary file "/home/runner/.cache/pocl/kcache/AH/MDDDDLEPEJELHLNMABPIBDIIJPEBOHCNCGCAN/getScale/64-1-1-goffs0-smallgrid/getScale.so", at _pocl_kernel_getScale_workgroup+0x1132 [0x7f6c884d8132]

...
/home/runner/work/_temp/e48d8ed4-04b5-4621-addf-7b9a2dc460e1.sh: line 3: 24673 Segmentation fault      (core dumped) pytest --verbose Test*
TestPythonTorchForce.py::TestPythonTorchForce::testMinimization 
```

It's crashing in a kernel inside the GPU minimizer.  Is it running with some CPU OpenCL implementation that's getting installed in the CI environment?  I can't reproduce it locally with either NVIDIA (GPU) or Intel CPU OpenCL; everything passes on my machine.

peastman:
I've been trying to reproduce it, also without success.  It's using pocl, which is a very buggy OpenCL.  That might or might not be the problem.

peastman:
After switching from pocl to Intel OpenCL (which is still buggy, but not quite as buggy), the test passes.  And I can see it running OpenCL tests, so it's not just that the platform gets skipped.

```
TestTorchForce.py::testForce[OpenCL-True-../../tests/central.pt-False-False] PASSED [ 73%]
TestTorchForce.py::testForce[OpenCL-True-../../tests/forces.pt-True-False] PASSED [ 75%]
TestTorchForce.py::testForce[OpenCL-True-../../tests/forces.pt-True-True] PASSED [ 77%]
```

epretti (approved):
Looks good.  I tried installing pocl on my machine and could reproduce the failure from before; the minimizer test and several others from the main OpenMM test suite failed with it, so I think it is just a broken implementation.  In any case, it seems not to be a problem with PythonTorchForce or the minimizer.

